PR #1, opened 2023-01-20
Adds a simple async JSON-RPC client that can make requests via HTTP and subscribe to notifications via WS

hmoniz:


src/jsonrpc/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+}
+
+#[async_trait]
+impl JsonRpcClient for JsonRpcClientImpl {
+    async fn request(&self, method: &str, params: Value) -> Result<Value> {
+        let request_body = build_jsonrpc_request(method, params)?;
+        let response = self.http_client.post(self.url.as_str())
+            .headers(HeaderMap::from_iter(
+                [(CONTENT_TYPE, HeaderValue::from_static("application/json"))]))
```
adlrocha, line 54: We will need to access operations from Lotus with an `admin/write` role, so we will have to attach the authentication token as part of the header in the `Bearer` field. See https://lotus.filecoin.io/reference/basics/api-access/#locally-hosted-node

src/jsonrpc/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+}
+
+// Processes a websocket stream by reading messages from the stream `ws_stream` and sending
+// them to an output channel `chan`.
+async fn handle_stream(mut ws_stream: WebSocketStream<MaybeTlsStream<TcpStream>>, chan: Sender<Value>) {
+    loop {
+        match ws_stream.next().await {
+            None => {
+                // TODO: does this mean that the channel was closed?
```
adlrocha, line 82: Correct. See https://docs.rs/ws_stream_tungstenite/0.1.0-alpha.4/ws_stream_tungstenite/#how-to-close-a-connection. Receiving a `None` when polling the steam signals that the websocket channel has been closed.

src/jsonrpc/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+        let (send_chan, recv_chan) = async_channel::unbounded::<Value>();
+        spawn(handle_stream(ws_stream, send_chan));
+
+        Ok(recv_chan)
+    }
+}
+
+// Processes a websocket stream by reading messages from the stream `ws_stream` and sending
+// them to an output channel `chan`.
+async fn handle_stream(mut ws_stream: WebSocketStream<MaybeTlsStream<TcpStream>>, chan: Sender<Value>) {
```
adlrocha, line 78: Definitely a future-work once we figure out the architecture of the client, but we may be able to reuse the websocket stream to also send new commands to Lotus without requiring an additional connection.

src/jsonrpc/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+            None => {
+                // TODO: does this mean that the channel was closed?
+                log::error!("No message in websocket stream");
+                break;
+            }
+            Some(result) => {
+                match result {
+                    Ok(msg) => {
+                        log::trace!("Read message from websocket stream: {}", msg);
+                        let value = serde_json::from_str(msg.to_text().unwrap()).unwrap();
```
adlrocha, line 90: Also future work: Instead of deserializing into a `serde_json::Value` we may want deserialize directly into the right type. This will be really convenient specially when we start inspecting the state from actors.

adlrocha (approved):
A few minor comments. The only one worth addressing right now is probably the inclusion of the `Bearer` header to authenticate with the `admin` role. Thanks!

src/jsonrpc/mod.rs
hmoniz: Absolutely. This is far from a closed design and we'll be evolving the interface as the use cases become more clear.

src/jsonrpc/mod.rs
hmoniz: Yep, agree. I'm thinking that that will be a layer on top of this interface and opted to keep this as simple and generic as possible.

aakoshh:
Possibly worth looking at https://github.com/ChainSafe/forest/tree/main/node/rpc-client

src/jsonrpc/mod.rs
hmoniz: Added support for the bearer token

hmoniz:
> 
We looked into it. Unfortunately, it doesn't support websockets, which we need.

PR #2, opened 2023-01-26
add skeleton

cryptoAtwill:
## Changes

 Add skeleton for ipc client cli tool.

- Add node cli to launch http server using `warp`
- Add health check cli
- Add basic config file parsing
- Added `cicd` and other dev linting tool, formatted previous json rpc code
- Add health check endpoint to client node

## Tests

```shell
cargo build --release -p client --bin ipc-client
# once built, launch with
./target/release/ipc-client

# test health check
./target/release/ipc-client node  # to start the node
./target/release/ipc-client healthcheck # to check the node health, it should print `node: ... is healthy` if node is up
```


client/src/bin/ipc-client.rs
```diff
@@ -0,0 +1,14 @@
+use client::{cli, health_check, start_node, HEALTH_CHECK, NODE};
+
+#[tokio::main]
+async fn main() {
+    env_logger::init();
+
+    let cmd = cli();
+    let matches = cmd.get_matches();
```
adlrocha, line 8: We should figure out a cleaner and more general way of doing this so including new commands is as easy as implementing some kind of `RpcCmd` trait and providing it to an enum of available commands.

client/src/command/health.rs
```diff
@@ -0,0 +1,45 @@
+use crate::config::{DEFAULT_NODE_ADDR, DEFAULT_PROTOCOL};
+use crate::node::{HealthResponse, DEFAULT_HEALTH_ENDPOINT};
+use clap::ArgMatches;
+use lazy_static::lazy_static;
+
+lazy_static! {
```
adlrocha, line 6: Can we set a default endpoint from which the node always listens (e.g. `127.0.0.1:5555`) and pick up the endpoint the client points to for the rpc calls either through an env variable (like `IPC_ENDPOINT_ADDR`) or a config file?

client/src/command/health.rs
```diff
@@ -0,0 +1,45 @@
...
+    static ref DEFAULT_URL: String = format!(
+        "{}://{}/{}",
+        DEFAULT_PROTOCOL, DEFAULT_NODE_ADDR, DEFAULT_HEALTH_ENDPOINT
+    );
+}
+
+pub(crate) const HEALTH_CHECK_ENDPOINT_KEY: &str = "NODE_ENDPOINT";
```
adlrocha, line 13: Let's build a trait for commands where you define a key, a set of expected args, the request and response, and a handler so we can implement them in a general way. Also we may need to support a chain of subcommands (to make calls such as `./ipc-client checkpoint list`

client/src/command/health.rs
```diff
@@ -0,0 +1,45 @@
...
+            log::error!("cannot query health endpoint: {:?} due to {:?}", node, e);
+            return false;
+        }
+        Ok(r) => r,
+    };
+
+    r.json::<HealthResponse>()
+        .await
+        .map(|n| n.is_healthy)
+        // this would be a parsing error, which we will treat as unhealthy
```
adlrocha, line 42: I don't think is going to be that rare, you may be using the `./ipc-client` and forgot to spawn the node or point to the right endpoint.

client/src/command/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+
+pub use health::health_check;
+pub use node::start_node;
+
+pub const NODE: &str = "node";
+pub const HEALTH_CHECK: &str = "healthcheck";
+
+pub fn cli() -> Command {
+    Command::new("ipc-client")
+        .about("The one client to interact with IPC actors")
```
adlrocha, line 17:
```suggestion
        .about("The one client to run IPC")
```

client/src/command/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+        .subcommand_required(true)
+        .arg_required_else_help(true)
+        .allow_external_subcommands(true)
+        .subcommand(
+            Command::new(HEALTH_CHECK)
```
adlrocha, line 22: Linked to my comments above, we should figure out a cleaner way to pass the subcommands and the handlers to the clients and servers.

client/src/config.rs
```diff
@@ -0,0 +1,43 @@
...
+use serde::Deserialize;
+use std::net::SocketAddr;
+use std::str::FromStr;
+
+pub(crate) const DEFAULT_NODE_ADDR: &str = "127.0.0.1:3030";
+pub(crate) const DEFAULT_PROTOCOL: &str = "http";
+
+/// The Client Node Configuration. Should be loaded from a static file.
+#[derive(Deserialize, Debug, Default)]
+pub struct ClientNodeConfig {
```
adlrocha, line 11: I think there should be a single config for the whole IPC client/node that is shared. We should think as the `IPC client` as a single application. Actually, calling it client may not be the best naming, and we should maybe call it `ipc-node`.

adlrocha (changes requested):
We need the code structure to be cleaner and general to make implementing new commands easier. Ideally, implementing a command should mean implementing a trait and providing it to an enum. This trait would provide the request, response, args, and handlers for the server (node) and the cli (interacting with the node through rpc). 

Also, I see that calling it `ipc-client` may lead to some confusion (as we implement a process orchestrating all IPC-specific mechanics). We should maybe call this application the `ipc-node`. Thoughts? 

(chiming @hmoniz in for input here)

adlrocha:
Would you mind adding a few lines in the `README` describing how to compile, run, etc.? I think it's easier if we populate the README with every new change instead of having to add all of it at once. Thanks :pray:

hmoniz:
> We need the code structure to be cleaner and general to make implementing new commands easier. Ideally, implementing a command should mean implementing a trait and providing it to an enum. This trait would provide the request, response, args, and handlers for the server (node) and the cli (interacting with the node through rpc).
> 
> Also, I see that calling it `ipc-client` may lead to some confusion (as we implement a process orchestrating all IPC-specific mechanics). We should maybe call this application the `ipc-node`. Thoughts?
> 
> (chiming @hmoniz in for input here)

I understand the issues you're facing regarding the UX and I'm trying to figure them out here in this doc: https://docs.google.com/document/d/14lkRRv6MQYnuEfp2GoGngdD8Q5YgfE38D8HTZWKgKf4/edit?usp=sharing

Since there is a bit of detail to figure out, I think we should nail down the design before investing too much into the code.

client/src/command/health.rs
cryptoAtwill: there is a config for this and variable defined

cryptoAtwill:
I think it would be quite time consuming to create a trait that captures all the information of a command, it would lead to some complex `macro` that takes time to tune. Since the project is still new for now and there will be changes, I feel it's enough if we can separate the functions into layers, like command line commands (api layer or controller layer), then api layer calls into the services (actual execution of the commands with the parameters). For the services we can have a trait and do some simple and quick macro to have dispatching. Also once we are sure of the pattern, then we can simplify based on the design.

client/src/cli/util.rs
```diff
@@ -0,0 +1,45 @@
...
+        enum Commands {
+            $(
+                $name(<$handler as CommandLineHandler>::Request),
+            )*
+
+        }
+
+        pub async fn cli() {
+            let args = IPCNode::parse();
+            let r = match &args.command {
```
adlrocha, line 30: Isn't this possible with an enum instead of a macro?

client/src/node.rs
```diff
@@ -0,0 +1,49 @@
...
+        let json_rpc = warp::post()
+            .and(warp::path("json_rpc"))
+            .and(warp::body::bytes())
+            .map(|bytes: bytes::Bytes| {
+                log::info!("received bytes = {:?}", bytes);
+                warp::reply::json(&())
+            });
+
+        let health = warp::get()
+            .and(warp::path(DEFAULT_HEALTH_ENDPOINT))
```
adlrocha, line 33: I was thinking that the same trait you use for the `cli` could be use to register the handler for the `json_rpc` server handler and the response and requests

adlrocha (commented):
Thank you for this next round of iteration, @cryptoAtwill. Is late Friday evening here, so bear with me if any of my comments don't make sense (I will review again on Monday). 

We are getting close but it's still a bit messy. I like how there's a file for each command (or kind of commands) that implement the trait, but this trait is currently only used for the cli. I was thinking for this trait to become the basic object loaded by the cli and the json-rpc server with all the needed information. We could bring the implementations of this trait to a `cmd` package and then the cli and warp rpc server would load the corresponding commands.

Also, I like the macro-approach to register the commands, but I feel we can make something work with just an enum without over-complicating it with macros (if there is no other option this works for me). In any case, let's sync on Monday and sketch a quick doc with the design before starting to code again (I feel bad for making you constantly change things). Feel free to take a first stab at the design before I wake up on Monday :pray:

client/src/cli/util.rs
cryptoAtwill: wouldn't a macro be easier to use? The declaration and routing is done behind the scene so that only one place is needed to be updated.

client/src/node.rs
cryptoAtwill: Oh, that might need to be more careful. Those 2 ought to be 2 traits? For cli, it is quite natural to have no states in the handler, but for `json_rpc`, the handler not only might have states but also have dependencies among each other. Also their parameters might be different as well. For example, `healthcheck` cli handler is just taking the url of which node to ping, but for `json_rpc`, it might be taking some other dependent handler or services. Might be difficult to share the same handle for both `cli` and `json_rpc`.

adlrocha: Sure, sure. The idea is to have a `CliHandler` and a `JsonRPCHandler`. Actually, you can even make them independent traits for each and use a type bound in the macro (or enum) for both, so they are loaded to the CLI and spawned in the server.

client/README.md
```diff
@@ -0,0 +1,8 @@
+# IPC-Actor Client
+The client to interact with IPC client.
+
+## Dev
```
adlrocha, line 4: We need to update this README,. right?

client/src/bin/ipc.rs
```diff
@@ -0,0 +1,25 @@
+use client::{register_cli_command, register_server_routes, HealthCheckHandler};
+use lazy_static::lazy_static;
+
+// initialize your server RPC handlers here
+lazy_static! {
+    static ref HEALTH: Arc<HealthCheckHandler> = Arc::new(HealthCheckHandler {});
```
adlrocha, line 6: Why does this need to be static?

client/src/bin/ipc.rs
```diff
@@ -0,0 +1,25 @@
...
+}
+
+// define the routes here
+register_server_routes!(
+    {get, "/health-check", HEALTH}
```
adlrocha, line 11:
```suggestion
    {post, "/healthcheck", HEALTH}
```

client/src/config.rs
```diff
@@ -0,0 +1,23 @@
+use serde::Deserialize;
+use std::net::SocketAddr;
+use std::str::FromStr;
+
+pub(crate) const DEFAULT_NODE_ADDR: &str = "127.0.0.1:3030";
+pub(crate) const DEFAULT_PROTOCOL: &str = "http";
+pub(crate) const DEFAULT_HEALTH_ENDPOINT: &str = "health";
```
adlrocha, line 7: We shouldn't keep command-related constants here

adlrocha (commented):
First quick pass, this looks better, but I have to tinker a bit with the code. I still feel we could use a simpler and cleaner approach (maybe I'm wrong). Also, we are not currently implementing a JSONRPC but an RPC over URI. We should maybe consider starting with a JSON-RPC, and then move to an URI-based RPC if needed. WDYT?

client/src/config.rs
cryptoAtwill: The are configs for the node url, the defualt settings, not for the command line

client/src/bin/ipc.rs
cryptoAtwill: just make it easier to initialize for now, I will move them to some initialization macro, maybe that's better

PR #3, opened 2023-02-01
Cleanup

cryptoAtwill:
## Changes

Clean up as follows:
- [x] Info logger by default for node.
- [x] Make the initialization of the node not in the macro but in the commands to be able to change the processes that we run.
- [ ] Enum register_server_routes! ideally should be an enum or vector.
- [x] JSONRPCParam should take the jsonrpc version from a config or a const (like we do with other configs).
- [x] Make CommandLineHandler return loggers for success and failures instead of println. Maybe instead of returning a Result<(), Error> we could automatically process the Result and output the loggers accordingly

`Enum register_server_routes! ideally should be an enum or vector.` is not done yet because there are some difficulties around integrating with futures and capturing variables in `wrap`. Added a todo in the code and will come up with ways to simplify this further down the road.
 
## Tests

```
cargo build
./target/debug/agent node
./target/debug/agent health-check
```

### DIFF
--- a/agent/src/main.rs
+++ b/agent/src/main.rs
@@ -15,6 +15,9 @@ register_server_routes!(
 
         (h1, h2)
     },
+    // health_check method is associated with h1
+    // health_check2 method is associated with h2
+    // TODO: find a clearer way to association and not by convention
     commands: health_check, health_check2
 );
 
--- a/agent/src/main.rs
+++ b/agent/src/main.rs
@@ -1,11 +1,14 @@
-use agent::command::health::HealthCheckCmd;
-use agent::{register_cli_command, register_server_routes};
+mod command;
+mod common;
+
+use crate::command::health::HealthCheckCmd;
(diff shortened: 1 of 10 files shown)


agent/src/main.rs
```diff
@@ -1,29 +1,37 @@
...
+        use crate::command::health::HealthCheckCmd;
 
         let h1 = HealthCheckCmd {};
         let h2 = HealthCheckCmd {};
 
         (h1, h2)
     },
+    // health_check method is associated with h1
+    // health_check2 method is associated with h2
+    // TODO: find a clearer way to association and not by convention
```
adlrocha, line 20: What do you mean by convention?

adlrocha (approved):
Yay! I think this pretty much covers it for now! Thank you!
I left one comment that I would like to clarify (what you mean by convention, and how are methods currently being loaded). But we can discuss tomorrow.

PR #8, opened 2023-02-14
A module for reading the agent config file

hmoniz:
This PR adds a new module `config` that provides a function `Config::from_file(path)` that reads a config TOML file for the agent and deserializes it in a type-safe way into a `Config` struct representing the config's contents.


config/test_example.toml
```diff
@@ -0,0 +1,12 @@
+[subnets]
+
+[subnets.root]
+path = "/root"
+rpc_api = "https://example.org/rpc/v0"
```
cryptoAtwill, line 5: shouldn't we call it `http_url` to be specific? Since the `JsonRpcClient` uses `ws`, if I pass in `ws`, will it work?

src/config.rs
```diff
@@ -0,0 +1,136 @@
...
+use ipc_sdk::subnet_id::SubnetID;
+use serde::de::{Error, SeqAccess};
+use serde::{Deserialize, Deserializer};
+use url::Url;
+
+/// The top-level struct representing the config. Calls to [`Config::from_file`] deserialize into this
+/// struct.
+#[derive(Deserialize)]
+pub(crate) struct Config {
+    pub subnets: HashMap<String, Subnet>,
```
cryptoAtwill, line 26: The key to the hashmap, is it a name? or it will only be `child` and `root` or it's actually the caller can call whatever it is?

What happens if the `subnets` had different key, but the Subnet values configurations is the same? Do we consider them to the be same or should we raise a warning or let caller be responsible?

src/config.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/// Represents a subnet declaration in the config.
+#[derive(Deserialize)]
+pub struct Subnet {
+    #[serde(deserialize_with = "deserialize_path")]
+    path: SubnetID,
```
cryptoAtwill, line 33: Why not just call it `subnet_id`? Could multiple subnets have the same`subnet_id`? If possible, I assume their `path` would be different, but they can have the same `subnet_id`? Actually what does the `path` mean, maybe attach the design doc could help.

src/config.rs
```diff
@@ -0,0 +1,136 @@
...
+    rpc_api: Url,
+    auth_token: Option<String>,
+    #[serde(deserialize_with = "deserialize_accounts", default)]
+    accounts: Vec<Address>,
```
cryptoAtwill, line 37: I think would be nice to add a link to the design doc, explaining what `accounts` stand for.

config/test_example.toml
```diff
@@ -0,0 +1,12 @@
...
+auth_token = "token_root"
+
+[subnets.child]
+path = "/root/f3thgjtvoi65yzdcoifgqh6utjbaod3ukidxrx34heu34d6avx6z7r5766t5jqt42a44ehzcnw3u5ehz47n42a"
```
adlrocha, line 9: nit: we are currently using `f0` addresses, for subnets (although now that I see this in the future we could also consider using `f3` addresses even if they are longer because they are reorg safe, I'll track it on an issue an issue). You can use `"/root/f0100"` as an example of a valid `subnetID`

src/config.rs
```diff
@@ -0,0 +1,136 @@
...
+    use fvm_shared::address::Address;
+    use ipc_sdk::subnet_id::SubnetID;
+    use url::Url;
+
+    use crate::config::Config;
+
+    #[test]
+    fn read_example_config() {
+        let path = "config/test_example.toml";
+        let test_addr = "f3thgjtvoi65yzdcoifgqh6utjbaod3ukidxrx34heu34d6avx6z7r5766t5jqt42a44ehzcnw3u5ehz47n42a";
```
adlrocha, line 114: You can also use `Address::new_id(1001)` to generate a valid `f0` address for the `SubnetID`

adlrocha (approved):
A small nit regarding the type of addresses that we use in `SubnetID`. Otherwise, it looks good. I like how you use `serde::Visitor` for zero-copy string serialization :)

src/config.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    use fvm_shared::address::Address;
+    use ipc_sdk::subnet_id::SubnetID;
+    use url::Url;
+
+    use crate::config::Config;
+
+    #[test]
+    fn read_example_config() {
+        let path = "config/test_example.toml";
```
cryptoAtwill, line 113: This path `config/test_example.toml` is relative, if the tests are triggered from other paths but the repo root, then tests will break. Using the file's absolute path as base would be safer for tests. But still, `test_example.toml` could potentially be removed, then tests will break as well. Or we can dynamically create temp files and destroy them, which will be neater.
I think we can just implement `From<String> for Config` and then Config::from_file() calls Config::from(string), and in tests, there is no need to depend on external files.

src/config.rs
```diff
@@ -0,0 +1,136 @@
+//! Provides a simple way of reading configuration files.
+//!
+//! Reads a TOML config file for the IPC Agent and deserializes it in a type-safe way into a
+//! [`Config`] struct. Usage:
+//!
+//! ```
+//! let c: Config = Config::from_file("config/example.toml");
```
cryptoAtwill, line 7: We can put this in doc tests are well.

cryptoAtwill (changes requested):

config/test_example.toml
hmoniz: That's a really good point. The current `JsonRpcClient` implementation uses HTTP for the `request` method and WS for the `subscribe` method. This is not ideal, particularly because each instance is associated with a specific URL. This means that each instance of the current client implementation can only actually call one of the two methods of the trait. This is far from ideal.

Here's what I suggest as a solution:
- The config file should have a field for HTTP and another for WS (e.g., `jsonrpc_api_http` and `jsonrpc_api_ws`. (I'll do this in the PR)
- Each implementation of the `JsonRpcClient` should use a single transport protocol. Instead of the current `JsonRpcClientImpl`, we should have a `HttpJsonRpcClient` and a `WsJsonRpcClient` implementation, both implementing the `JsonRpcClient` trait. (I'll do this in a follow-up PR)

config/test_example.toml
hmoniz: Thanks! I'll make that change.

src/config.rs
hmoniz: Yes, I think that would be preferable. Perhaps, using simply `id` because the field is already under a `subnet` section in the config. @adlrocha What do you think of the name change?

src/config.rs
hmoniz: I get your point. However, I think we should be adding a link to the design doc in a README.md file. That's because we'll have these domain-specific terms spread out throughout the codebase and we don't want to add a link to each one of them. Wdyt?

cryptoAtwill: Yep, that works!

src/config.rs
hmoniz: Yes, thank you for bringing that up. I like this idea and it would make the test much more hermetic. The only downside is we won't be calling `Config::from_file` in a test, but I suppose we can live with that.

src/config.rs
hmoniz: The key to the hashmap comes from the configuration. If we have two subnets in the config like this:
```
[subnets]
[subnets.root]
# root subnet config
[subnets.child]
# child subnet config
```
Then the hashmap will have keys "root" and "child".

src/config.rs
hmoniz: The README will be added in a follow-up PR once we have more meat about the ux the agent.

cryptoAtwill (approved):

src/config.rs
hmoniz: Opted to remove since this example become obsolete with the removal of the file from the source tree.

PR #9, opened 2023-02-15
Add `warp` based Http JSON RPC node

cryptoAtwill:
This pr adds the JSON RPC node based on `warp`.

# Examples
```rust
use agent::node::config::IPCAgentJsonRPCNodeConfig;
use agent::node::jsonrpc::IPCAgentJsonRPCNode;

#[tokio::main]
async fn main() {
    let n = IPCAgentJsonRPCNode::new(IPCAgentJsonRPCNodeConfig::default());
    n.run().await;
}
```

The following PRs are planned as such:
1. Add in the `IPCAgentJsonRPCHandler` trait that will be shared by all the RPC handlers.
2. Auto routing of `method` and `params` to the Handler.

Btw, my local cargo pipeline will always format the existing code. We should do another PR to add in cicd checks and format code one shot, #10 .


src/jsonrpc/mod.rs
```diff
@@ -42,6 +42,7 @@ pub struct JsonRpcClientImpl {
 
 impl JsonRpcClientImpl {
     /// Creates a client that sends all requests to `url`.
+    #[allow(dead_code)]
```
adlrocha, line 45: This `dead_code` directive should not be needed once you rebase from master, right? The `LotusClient` makes use of the `JsonRpcClientImpl` IIRC, right?

src/node/config.rs
```diff
@@ -0,0 +1,22 @@
+use serde::Deserialize;
+use std::net::SocketAddr;
+use std::str::FromStr;
+
+pub const DEFAULT_JSON_RPC_ADDR: &str = "127.0.0.1:3030";
+pub const DEFAULT_JSON_RPC_PROTOCOL: &str = "http";
```
adlrocha, line 6: We should make these configurable either through the IPC agent config file, or through a CLI flag when we start the IPC agent deamon. I guess it makes more sense to adress it right away in a follow-up PR. Can you add an issue or prepare a PR to tackle this? Thanks!

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,149 @@
...
+}
+
+impl IPCAgentJsonRPCNode {
+    pub fn new(config: IPCAgentJsonRPCNodeConfig) -> Self {
+        Self { config }
+    }
+
+    /// Runs the node in the current thread
+    pub async fn run(&self) {
+        log::info!("rpc node started at {:?}", self.config.addr());
```
adlrocha, line 36:
```suggestion
        log::info!("IPC agent rpc node listening at {:?}", self.config.addr());
```

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,149 @@
...
+/// Note that currently only http json rpc is supported.
+///
+/// # Examples
+/// ```no_run
+/// use agent::node::config::IPCAgentJsonRPCNodeConfig;
+/// use agent::node::jsonrpc::IPCAgentJsonRPCNode;
+///
+/// #[tokio::main]
+/// async fn main() {
+///     let n = IPCAgentJsonRPCNode::new(IPCAgentJsonRPCNodeConfig::default());
```
adlrocha, line 21: `addr` is an `Option` in the config, are we sure that this works? If I am not mistaken the default for an `Option` is `None`. I don't know the behavior of `run` over a `None` address, but we should probably figure out a better way to pass a default configuration.

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// Runs the node in the current thread
+    pub async fn run(&self) {
+        log::info!("rpc node started at {:?}", self.config.addr());
+        warp::serve(json_rpc_filter()).run(self.config.addr()).await;
+    }
+}
+
+// Internal implementations
+
+/// Create the json_rpc filter
```
adlrocha, line 43: Can you document what this filter does: 
- Listen to POST requests on the `DEFAULT_JSON_RPC_ENDPOINT` 
- Extract the body of the request. 
- Pass it to to the process function.

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub method: String,
+    pub params: Value,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct JSONRPCResponse<T: Serialize> {
+    pub id: u16,
+    pub jsonrpc: String,
+    pub result: T,
+}
```
adlrocha, line 98: Can we extract these to a `types.rs` mod?

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,149 @@
...
+/// Create the json_rpc filter
+fn json_rpc_filter() -> impl Filter<Extract = (warp::reply::Json,), Error = warp::Rejection> + Copy
+{
+    warp::post()
+        .and(warp::path(DEFAULT_JSON_RPC_ENDPOINT))
+        .and(warp::body::bytes())
+        .and_then(process)
+}
+
+/// The process method for the JSON RPC node.
```
adlrocha, line 52: Same here, it would be great to have a better description of what this function does:
- Route the request to the right handler.
- Send request.

Also, what about calling it `handle_request` instead of `process`?

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,149 @@
...
+    log::debug!("received bytes = {:?}", bytes);
+
+    match serde_json::from_slice::<JSONRPCParam>(bytes.as_ref()) {
+        Ok(p) => {
+            let JSONRPCParam {
+                id,
+                method,
+                params,
+                jsonrpc,
+            } = p;
```
adlrocha, line 63: Why can't the deserialization into json be part of the filter (the function above), and then this function exclusively handle the routing/handling of functions? 

I would move this logic of deserializing the body into json up to the filter, and I would make this function be the one that picks up the request and receives the request and tailors the response.

You can add a placeholder here or a sample `echo` handler here to showcase how that would be structured.

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,149 @@
...
+            Ok(warp::reply::json(&JSONRPCResponse {
+                id: 0,
+                jsonrpc: String::from(DEFAULT_JSON_RPC_VERSION),
+                result: serde_json::Value::String(String::from("Cannot parse parameters")),
+            }))
+        }
+    }
+}
+
+/// Follows: https://ethereum.org/en/developers/docs/apis/json-rpc/#curl-examples
```
adlrocha, line 84: Can we have a bit more of documentation here? Say that this are the standard types for our json-rpc and that follow a structure similar to the one of the Ethereum RPC.

adlrocha (changes requested):
Way more simple than the initial approach, thank you @cryptoAtwill. A few suggestions here and there but we are in the right path.

src/node/jsonrpc.rs
cryptoAtwill: reason I put it here because in the scope of the current PR, there is no other struct or module using this. It's completely internal to this file. Maybe I will move it to a separate file when there is a use case for this?

src/node/config.rs
cryptoAtwill: In the configuration, we have a filed to handle the user input. I was thinking the default should be when user has not input anything. If this is from config, then the question is what happens if user does not pass in those default addr or protocol? Then we are asking them to define both the default and the expected url?

src/node/jsonrpc.rs
cryptoAtwill: in the code, it's calling `config.addr()` which handles the `None` and convert to the address to a predefined socket addr.

src/node/jsonrpc.rs
cryptoAtwill: ok

src/node/jsonrpc.rs
cryptoAtwill: merged

src/node/jsonrpc.rs
cryptoAtwill: The main reason is because this way I dont really have to write another warp middleware, which should terminate the execution if deserialization fails. I was thinking of handling the routing by creating another method in this function body, will that work?

src/jsonrpc/mod.rs
cryptoAtwill: let me try

src/node/jsonrpc.rs
adlrocha: Gotcha, so they are not publicly exposed yet. I was expecting them to be in the future once the CLI and external libraries need these types to perform jsonrpc requests. I am fine leaving them here for now then, nonetheless can you move them then to the beginning of the file, before the functions declarations so we have first the type and trait definitions and then the functions?

src/node/jsonrpc.rs
adlrocha: Please, remember to resolve the threads once the comment has been address (that way I know where to look when I do the second pass to the review) 🙏

src/node/jsonrpc.rs
adlrocha: I think is cleaner to have the deserialization on the top filter and the routing in this function. That way we have a clear separation of roles. Why can't you terminate the execution in the execution if the deserialization fails?

src/node/jsonrpc.rs
cryptoAtwill: The other way I considered was actually in serde default deserialization, but that would result in losing the `#[derive(Default)]` and one cannot simply do:
```
IPCAgentJsonRPCNodeConfig{ addr: ..., ..Default::default() }
```
Any other methods you have in mind?

src/node/jsonrpc.rs
cryptoAtwill: updated and resolve this.

src/node/jsonrpc.rs
cryptoAtwill: Nah, just need to write more code which probably it's a style preference and might not have liked. I have updated the code accordingly, check it out:)

src/node/jsonrpc.rs
cryptoAtwill: updated and resolving this

src/node/jsonrpc.rs
cryptoAtwill: updated and resolving this.

src/node/jsonrpc.rs
cryptoAtwill: updated and resolving this.

src/node/config.rs
adlrocha: In the design doc we considered that the daemon always accepts the path of a config file (or use the default config location) as input: https://docs.google.com/document/d/14lkRRv6MQYnuEfp2GoGngdD8Q5YgfE38D8HTZWKgKf4/edit#heading=h.i71qsbf2iwn5

With this in mind I don't think the default address will ever be not set, as a config file should be picked up, even if it is the default one. In any case, I guess we can address this in the coming PR with the daemon command and discuss how to handle it there. WDYT?

src/node/jsonrpc.rs
adlrocha: Oh, I see... so `addr` is private and when accessing it through `addr()` it auto-assigns the default value. The thing that I am now wondering, is if once we force the use of a config file as input we need `addr` to be an `Option` at all, but I guess we can discuss that in the next iteration.

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,172 @@
...
+        warp::serve(json_rpc_filter()).run(self.config.addr()).await;
+    }
+}
+
+// Internal implementations
+
+/// Create the json_rpc filter. The filter does the following:
+/// - Listen to POST requests on the DEFAULT_JSON_RPC_ENDPOINT
+/// - Extract the body of the request.
+/// - Pass it to to the process function.
```
adlrocha, line 66:
```suggestion
/// - Pass it to to the json_rpc_filter to deserialize into a jsonrpc request.
```

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+async fn to_json_rpc_request(bytes: Bytes) -> Result<JSONRPCRequest, warp::Rejection> {
+    serde_json::from_slice::<JSONRPCRequest>(bytes.as_ref()).map_err(|e| {
+        log::debug!("cannot deserialize {bytes:?} due to {e:?}");
+        warp::reject::custom(InvalidParameter)
+    })
+}
+
+/// To handle the json rpc request. Currently just log it.
```
adlrocha, line 83:
```suggestion
/// Main function responsible for handling and routing jsonrpc requests to the right underlying handler according to the method
```

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,172 @@
...
+/// - Pass it to to the process function.
+fn json_rpc_filter() -> impl Filter<Extract = (impl Reply,), Error = warp::Rejection> + Copy {
+    warp::post()
+        .and(warp::path(DEFAULT_JSON_RPC_ENDPOINT))
+        .and(warp::body::bytes())
+        .and_then(to_json_rpc_request)
+        .and_then(handle_request)
+        .recover(handle_rejection)
+}
+
```
adlrocha, line 75:
```suggestion
// Filter that deserializes the body of the request into a jsonrpc request.
```

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,172 @@
...
+        let jsonrpc = String::from(DEFAULT_JSON_RPC_VERSION);
+        let id = 0;
+
+        let req = JSONRPCRequest {
+            id,
+            jsonrpc: jsonrpc.clone(),
+            method: foo.clone(),
+            params: Default::default(),
+        };
+        // Execute `sum` and get the `Extract` back.
```
adlrocha, line 143: Is this command correct?

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,172 @@
...
+        assert_eq!(v.id, id);
+        assert_eq!(v.jsonrpc, jsonrpc);
+        assert_eq!(v.result, ());
+    }
+
+    #[tokio::test]
+    async fn test_json_rpc_filter_cannot_parse_param() {
+        let filter = json_rpc_filter();
+
+        // Execute `sum` and get the `Extract` back.
```
adlrocha, line 162: Same here

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        let v = serde_json::from_slice::<JSONRPCResponse<()>>(value.body()).unwrap();
+
+        assert_eq!(v.id, id);
+        assert_eq!(v.jsonrpc, jsonrpc);
+        assert_eq!(v.result, ());
+    }
+
+    #[tokio::test]
+    async fn test_json_rpc_filter_cannot_parse_param() {
```
adlrocha, line 159: If you see that it is not really hard, would you mind adding some quick tests with the error cases for `handle_rejection`? Just in case we introduce regressions in the future. Thanks!

adlrocha (approved):
A new round of light suggestions, but overall I think it is in a good state now. Thanks. Summoning @hmoniz in case he wants to do a quick pass before merging.

src/node/config.rs
cryptoAtwill: Yeah, I can use that config file. But in the file you sent, I dont see a config parameter for the addr of the current node? Or it's not added yet?

src/node/jsonrpc.rs
cryptoAtwill: Yep, make sense. If we are using the config file, I dont even need this config struct. I can use that config file. But just need to define a field for this addr.

src/node/jsonrpc.rs
cryptoAtwill: yep, should be quick. Will add it.

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,172 @@
...
+        log::error!("unhandled rejection: {:?}", err);
+        Ok(with_status(
+            "INTERNAL_SERVER_ERROR",
+            StatusCode::INTERNAL_SERVER_ERROR,
+        ))
+    }
+}
+
+#[cfg(test)]
+mod test {
```
hmoniz, line 124: nit: the convention is to the plural form, i.e., `mod tests`.

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,172 @@
...
+    pub id: u16,
+    pub jsonrpc: String,
+    pub method: String,
+    pub params: Value,
+}
+
+/// The json rpc response. It is the standard form our json-rpc and follows a structure similar
+/// to the one of the Ethereum RPC: https://ethereum.org/en/developers/docs/apis/json-rpc/#curl-examples
+#[derive(Debug, Serialize, Deserialize)]
+struct JSONRPCResponse<T: Serialize> {
```
hmoniz, line 23: This must also include an `error` field to conform with the JSON-RPC spec (see https://www.jsonrpc.org/specification#response_object). The serialization then uses one of `result` and `error` depending on the outcome of the processing of the request.

hmoniz (changes requested):
This PR is definitely on a good trajectory. I left a few comments. The most important ones are related to the config and the JSON-RPC response missing the error field.

I also suggest a bit of a code restructuring, which perhaps is better addressed in a follow-up PR. The suggestion is the following: Let's have a `jsonrpc` module with submodules `client` and `server` (or `node` - I have a slight preference for `server` but curious to hear others thoughts).

src/node/jsonrpc.rs
cryptoAtwill: right right, then in this case I think we can do something like this:
```
struct JSONRPCResponse<T: Serialize> {
    pub id: u16,
    pub jsonrpc: String,
    #[some serde config to ignore serialization if None]
    pub result: Option<T>,
    #[some serde config to ignore serialization if None]
    pub error: Option<String>, // we can use T is we want support struct.
}
```
Then, we can based to `Result` to set `result` or `error`.

src/node/jsonrpc.rs
```diff
@@ -0,0 +1,172 @@
...
+use serde_json::Value;
+use warp::http::StatusCode;
+use warp::reject::Reject;
+use warp::reply::with_status;
+use warp::{Filter, Rejection, Reply};
+
+/// The json rpc request param. It is the standard form our json-rpc and follows a structure similar
+/// to the one of the Ethereum RPC: https://ethereum.org/en/developers/docs/apis/json-rpc/#curl-examples
+#[derive(Serialize, Deserialize, Debug)]
+struct JSONRPCRequest {
```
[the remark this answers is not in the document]
cryptoAtwill: Yep yep, I will do this in a follow up PR, seems it can be shared and maybe placed in to `util` or `common` module or we can just import from `jsonrpc` mod?

src/jsonrpc/mod.rs
```diff
@@ -199,4 +199,4 @@ fn build_jsonrpc_request(method: &str, params: Value) -> Result<Value> {
         })
     };
     Ok(request_value)
-}
+}
```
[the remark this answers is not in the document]
cryptoAtwill: yep yep, quite strange though, I don't see the changes locally. Let me check how did the formatting slipped in.

src/node/config.rs
adlrocha: Yes, that's a good point, @hmoniz. Let's call it `server` or if we want to be verbose (and just in case we end up with more future servers) we can use something like `AGENT_RPC_SERVER`

src/node/jsonrpc.rs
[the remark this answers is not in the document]
adlrocha: This goes in the line of a previous comment I made, if it is not a lot of work let's just move it right away instead of waiting for a follow-up PR. WDYT?

adlrocha (commented):
Thank you folks for the review and addressing the comments. @cryptoAtwill, let's resolve the few pending threads, fix merge conflicts and pipeline it for merge :)

src/node/jsonrpc.rs
cryptoAtwill: oops, you got me. Copy pasted sample from their official doc. Will remove.

src/node/config.rs
cryptoAtwill: Here is what I'm proposing:
Update existing `config.rs` into a mod, then the `Config` struct will stay in the `mod.rs`. The overall folder struct is something like:
```
config/
-- mod.rs
-- subnet.rs  // contains subnet related configs
-- server.rs  // contains server related configs
-- utils.rs  // contains all of the `deserialize_*` functions that can be shared.
```
This way maybe we can add more configs without problem. The other way which I was thinking of is actually putting the configs into each module, i.e. `server.rs` is placed in the module together with rpc node, so that all the code related to rpc node is contained in one folder. Both should be quick changes, I guess it's just a preference, maybe we try putting the configs in one folder first? 

@hmoniz @adlrocha any thoughts?

src/node/jsonrpc.rs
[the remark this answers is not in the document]
cryptoAtwill: The current `jsonrpc` mod contains the request and response, which was not merged when this code was written. Since those are merged, I can just use those directly.

src/node/jsonrpc.rs
cryptoAtwill: Update: 2023/02/20:
I think it's better not to mix the result and error. I read the [json rpc spec](https://www.jsonrpc.org/specification#response_object) of json_rpc, seems the error can be more complex. It's neater to use another struct just to handle the error.

src/node/config.rs
hmoniz: I really like your first suggestion as it keeps all the config (1) self-contained and (2) neatly modularized.

adlrocha: I like that approach of having a single config crate with topical files. I'll review the linked PR also. 🙏

src/node/jsonrpc.rs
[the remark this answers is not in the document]
adlrocha: I am fine either way. Whatever works best for you. Whenever this happens feel free to stack a new PR with the proposed change and link it from here so we can have the overview of the future changes and so you are not blocked by this being merged as you did with the config change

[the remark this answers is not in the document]
hmoniz: +1 for doing this in a separate PR

src/node/jsonrpc.rs
hmoniz: You can simply have here a field like: 
```
pub error: Option<JsonRpcError>
```
Is this what you're suggesting?

cryptoAtwill: Refer to this [struct](https://github.com/consensus-shipyard/ipc-agent/blob/1187b226e389d5323c8edc3f02f506534f3a683c/src/server/response.rs#L42). The usage is [here](https://github.com/consensus-shipyard/ipc-agent/blob/1187b226e389d5323c8edc3f02f506534f3a683c/src/server/jsonrpc.rs#L76)

src/node/jsonrpc.rs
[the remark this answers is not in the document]
hmoniz: BTW, it might be a good idea to address that PR first and then come back to this one.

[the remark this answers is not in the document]
cryptoAtwill: I think if it's not a big issue, maybe get this one merged first? Too many stacking up PRs will create quite some conflicts.

src/node/jsonrpc.rs
cryptoAtwill: As discussed, will close this one first then sharing the request and responses. Resolving first.

src/node/jsonrpc.rs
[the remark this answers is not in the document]
cryptoAtwill: I have create this issue to track it: #24, resolving this for now.

cryptoAtwill:
Hi @hmoniz @adlrocha, please help take a look since the config PR is merged. 

There is one outstanding issue, which is the shared Json RPC response and request. Based on previous discussion, I will do it in a follow up PR which is tracked in #24.

I think there are no other pending issues (maybe I missed something? lol). Please help take a look, thank you very much!

src/config/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+pub use crate::config::subnet::Subnets;
+
+pub const JSON_RPC_VERSION: &str = "2.0";
+
+/// The top-level struct representing the config. Calls to [`Config::from_file`] deserialize into
+/// this struct.
+#[derive(Deserialize)]
+pub(crate) struct Config {
+    pub server: Server,
+    pub subnets: Subnets,
```
hmoniz, line 26: nit: As mentioned in the other PR, I'd rather see here a `HashMap<String, Subnet>` instead of this custom type.

src/config/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+    const SERVER_JSON_RPC_ADDR: &str = "127.0.0.1:3030";
+    const ROOT_ID: &str = "/root";
+    const CHILD_ID: &str = "/root/f0100";
+    const ROOT_AUTH_TOKEN: &str = "ROOT_AUTH_TOKEN";
+    const CHILD_AUTH_TOKEN: &str = "CHILD_AUTH_TOKEN";
+    const JSONRPC_API_HTTP: &str = "https://example.org/rpc/v0";
+    const JSONRPC_API_WS: &str = "ws://example.org/rpc/v0";
+    const ACCOUNT_ADDRESS: &str = "f3thgjtvoi65yzdcoifgqh6utjbaod3ukidxrx34heu34d6avx6z7r5766t5jqt42a44ehzcnw3u5ehz47n42a";
+
+    fn check_server_config(config: &Server) {
```
hmoniz, line 66: I think it would be preferable to have this function and `check_subnets_config` as `#[test]` instead of `read_config`. This way, if there are issues with a specific config section, we know immediately which one is it by looking at the test that is failing.

hmoniz (approved):
Great job here. LGTM. I just left a couple of suggestions that don't need another review round.

src/config/mod.rs
cryptoAtwill: my bad, I didn't see this. Will revert this to `HashMap`.

src/config/mod.rs
adlrocha: Agree! Let's address this one and we should be good to go :pray:

adlrocha (approved):

src/config/mod.rs
cryptoAtwill: Cool, this is updated! Thx @adlrocha for the tip. Resolving this.

PR #11, opened 2023-02-16
Add cli command handler interface

cryptoAtwill:
This PR is the start of the series to introduce Command Line handling. The overall idea is, when adding a new command, one just needs to implement the core execution logic. The parsing of command line args, dispatching and response display are already taken care of. One just needs to register the handler to a registry. 

This PR introduces the trait that represents the abstraction of a command line handler.  To implement a new command line operation, implement this trait.

Example code can be found in #12.

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -15,6 +15,6 @@ pub trait CommandLineHandler {
     /// implementation to abstract away external crates. But this should be good for now.
     type Arguments: std::fmt::Debug + Args;
 
-    /// Handles the request and produces a response string.
+    /// Handles the request with the provided arguments. Dev should handle the content to print and how
     async fn handle(arguments: &Self::Arguments) -> anyhow::Result<()>;
 }
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -10,13 +10,13 @@ use clap::Args;
 /// constructed from scratch. Initialize the states in the `handle` method.
 #[async_trait]
 pub trait CommandLineHandler {
-    /// The request to process.
+    /// The command line argument to process.
     ///
     /// NOTE that this parameter is used to generate the command line arguments.
     /// Currently we are directly integrating with `clap` crate. In the future we can use our own
     /// implementation to abstract away external crates. But this should be good for now.
-    type Request: std::fmt::Debug + Args;
+    type Argument: std::fmt::Debug + Args;
 
     /// Handles the request and produces a response string.
-    async fn handle(request: &Self::Request) -> Result<String, Error>;
+    async fn handle(request: &Self::Argument) -> Result<String, Error>;
 }
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -7,8 +7,7 @@ use clap::Args;
 /// how to register.
 ///
 /// Note that this trait does not support a stateful implementation as we assume CLI commands are all
-/// constructed from scratch. Initialize the states in the `handle` method or pass in from the
-/// implementation constructor.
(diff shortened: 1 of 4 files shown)


src/cli/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+/// Note that this trait does not support a stateful implementation as we assume CLI commands are all
+/// constructed from scratch. Initialize the states in the `handle` method.
+#[async_trait]
+pub trait CommandLineHandler {
+    /// The request to process.
+    ///
+    /// NOTE that this parameter is used to generate the command line arguments.
+    /// Currently we are directly integrating with `clap` crate. In the future we can use our own
+    /// implementation to abstract away external crates. But this should be good for now.
+    type Request: std::fmt::Debug + Args;
```
adlrocha, line 18: Let's change `Request` to `Argument` so it is clear that the expected type is the argument for the cli command. Please, also update the comment to clarify that what we expect here is the set of commands supported by the cli command.

src/cli/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+
+    /// Handles the request and produces a response string.
+    async fn handle(request: &Self::Request) -> Result<String, Error>;
```
adlrocha, line 21: Let's maybe rename it to `handler` or `cmd_handler` and rename the argument to `args`. 

What do you think about also making an alias `type CmdOutput = String` to make it explicit and make it easier to change in the future?

src/cli/mod.rs
```diff
@@ -0,0 +1,22 @@
+use crate::error::Error;
+use async_trait::async_trait;
+use clap::Args;
+
+/// The trait that represents the abstraction of a command line handler. To implement a new command
+/// line operation, implement this trait and register it, see `register::register_cli_command` for
+/// how to register.
+///
+/// Note that this trait does not support a stateful implementation as we assume CLI commands are all
+/// constructed from scratch. Initialize the states in the `handle` method.
```
adlrocha, line 10:
```suggestion
/// constructed from scratch.
```

src/cli/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+#[async_trait]
+pub trait CommandLineHandler {
+    /// The request to process.
```
adlrocha, line 13:
```suggestion
    /// Abstraction for command line operations.
```
Let's try to avoid the name "request", let's use command line, arguments or method.

adlrocha (changes requested):
A few minor suggestions, but this looks fine. Thanks!

src/cli/mod.rs
cryptoAtwill: Previously we did use `type CmdOutput`, but decided to use just `String`. The reason for change is to enforce a printable message to the command line console display.

cryptoAtwill: clarified on slack and updated accordingly.

cryptoAtwill:
@adlrocha I have updated accordingly, please help take a look again.

src/cli/mod.rs
cryptoAtwill: Updated

src/cli/mod.rs
adlrocha: We agreed to return `anyhow::Result<()>` for now and let developers decide what to print and how.

adlrocha (approved):
LGTM, let's wait in case @hmoniz wants to give some additional input. In the meantime feel free to work on the follow-up. Thanks!

hmoniz (approved):
LGTM!

cryptoAtwill:
woohoo, merging!

PR #12, opened 2023-02-16
Sample Info command line

cryptoAtwill:
Sample `Info` command line handler to see how to work with `CommandLineHandler` trait, see #11 .


src/cli/commands/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+enum Commands {
+    Info(<info::Info as CommandLineHandler>::Arguments),
+}
+
+/// The overall command line struct to be used by `clap`.
+#[derive(Debug, Parser)]
+#[command(
+    name = "ipc",
+    about = "The IPC agent command line tool",
+    version = "v0.0.1"
```
adlrocha, line 62: Can we get the version from the github tag or maybe the current commit so we don't have to update it manually?

src/cli/commands/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+            e
+        )
+    }
+}
+
+/// The collection of all subcommands to be called, see clap's documentation for usage. Internal
+/// to the current mode. Register a new command accordingly.
+#[derive(Debug, Subcommand)]
+enum Commands {
+    Info(<info::Info as CommandLineHandler>::Arguments),
```
adlrocha, line 54: Out of a curiosity, why do we need to pass the `Arguments` here? I haven't looked in depth how `clap` works yet, so bear with me here :)

adlrocha (commented):
Looking good. As mentioned in Slack, let's keep this PR stale for now to illustrate how to add new commands and then decide if it makes sense to add this `info` command (right now it does nothing, so let's just keep it here for illustration purposes and decide if to merge or close). Thanks!

src/cli/commands/mod.rs
cryptoAtwill: `clap` requires the command line args to be passed in the enum, so technically, we need this:
```rust
enum Commands {
    Info(InfoArgs).
}
```
But since we have `InfoArgs` as an associate type to `info::Info`, we just need `info::Info as CommandLineHandler>::Arguments`. Maybe `InfoArgs` is more intuitive, but again, it's just to save some extra import and have some potential room to reduce code written.

src/cli/commands/mod.rs
adlrocha: This can probably be picked up in compile time by taking the current `git tag`. Let's explore it in the daemon cmd PR

PR #22, opened 2023-02-20
Json rpc server config

cryptoAtwill:
This is an example PR illustrating how config for `json_rpc` server and other mod will be refactored.


src/server/jsonrpc.rs
```diff
@@ -15,12 +15,17 @@ use warp::{Filter, Rejection, Reply};
...
 ///
+/// use std::net::SocketAddr;
+///
 /// #[tokio::main]
 /// async fn main() {
-///     let n = JsonRPCServer::new(JsonRPCServerConfig::default());
+///     let config = JsonRPCServerConfig {
+///         json_rpc_address: SocketAddr::from_str(server_json_rpc_addr).unwrap()
+///     }
+///     let n = JsonRPCServer::new(config);
```
adlrocha, line 28: Wouldn't it make more sense to share an example here that reads the config for the json rpc server from the config file instead of manually constructing the server config from an address variable? (is should be just a few more lines of code in the example).

src/server/mod.rs
```diff
@@ -1,11 +1,8 @@
...
 pub mod jsonrpc;
-mod request;
-mod response;
+pub mod request;
+pub mod response;
 
 const DEFAULT_JSON_RPC_SERVER_VERSION: &str = "2.0";
-pub const DEFAULT_JSON_RPC_SERVER_ENDPOINT: &str = "json_rpc";
-pub const DEFAULT_JSON_RPC_SERVER_PROTOCOL: &str = "http";
-pub const DEFAULT_JSON_RPC_SERVER_ADDR: &str = "127.0.0.1:3030";
```
adlrocha, line 11: Where these variables moved somewhere?

src/server/mod.rs
```diff
@@ -1,11 +1,8 @@
...
+const DEFAULT_JSON_RPC_SERVER_ENDPOINT: &str = "json_rpc";
```
adlrocha, line 8: As mentioned above, I would probably move these to the server config.

src/config/mod.rs
```diff
@@ -3,29 +3,36 @@
...
 }
 
+#[derive(Deserialize, Clone)]
+pub struct Server {
+    pub json_rpc_address: SocketAddr,
+}
+
 /// Represents a subnet declaration in the config.
 #[derive(Deserialize)]
 pub struct Subnet {
```
adlrocha, line 34: What about separating these into a `subnet.rs` and `server.rs` and move all relevant constants for them to these files? Like for instance `DEFAULT_JSON_RPC_SERVER_ENDPOINT`

adlrocha (changes requested):
First round of comments, I like this approach. I would suggest modularizing the config module into topical files (as I expect more config parameters to be added in the future). Let me know what you think 🙏

src/config/mod.rs
cryptoAtwill: I did that initially, but then I realized they are actually quite small. I was thinking of breaking into different files when they grow larger.

src/server/jsonrpc.rs
cryptoAtwill: works too!

src/server/mod.rs
cryptoAtwill: They are no longer needed, I removed them.

src/config/mod.rs
hmoniz: +1 to separating these into their own modules now.

src/config/mod.rs
```diff
@@ -49,72 +56,23 @@ impl Config {
...
     use fvm_shared::address::Address;
     use indoc::formatdoc;
-    use ipc_sdk::subnet_id::{SubnetID, ROOTNET_ID};
+    use ipc_sdk::subnet_id::{ROOTNET_ID, SubnetID};
     use url::Url;
 
     use crate::config::Config;
 
     #[test]
     fn read_config() {
```
hmoniz, line 72: Instead of a single monolithic test method, I suggest have separate test methods for the different config section. For example, keep the existing method and rename it to `read_config_subnets` and add a new method called `read_config_server` to test the server-specific logic.

src/server/mod.rs
hmoniz: I think these should be moved to the relevant module in the config crate or removed altogether. The latter option actually might make more sense because (1) we might want to force the user to be explicit about the server configuration and (2) we might not want to launch the server at all in some circumstances.

hmoniz (changes requested):

src/config/mod.rs
cryptoAtwill: @hmoniz This might have a problem as the fields must be present, if separate into two methods, both the `subnet` and `server` needs to be in the raw string, which seems to be strange.

src/server/mod.rs
cryptoAtwill: @hmoniz For `DEFAULT_JSON_RPC_SERVER_VERSION` it might be better to hard code as a const. This is actually a strict requirement from the spec. Maybe I call it `JSON_RPC_VERSION`, make it sound less like config. `DEFAULT_JSON_RPC_SERVER_ENDPOINT` I can make it into a config parameter.

cryptoAtwill: BTW, maybe we can just hard code the `json_rpc` endpoint?

src/server/mod.rs
cryptoAtwill: I have moved them already

src/server/mod.rs
cryptoAtwill: Just different host and port? They call share the same `json_rpc` path right? Like `localhost:3030/json_rpc` and `localhost:3031/json_rpc`, but the `/json_rpc` part remains the same.

src/config/mod.rs
cryptoAtwill: At least based on the current implementation, it is not possible to separate into sub tests. The other way to work around this is having:
```rust
pub struct Config {
    pub server: Option<Server>,
    pub subnets: Option<HashMap<String, Subnet>>,
}
```
Then whoever needs the specific section, just do `let server_config = config.server.expect("config not defined")`. Then we can achieve the above. 

@hmoniz @adlrocha This approach ok?

adlrocha: I am not a fan of using an `Option` if we want the config to be mandatory, it may lead to future issues. Something we could do is to have a single config template for the whole test mode and all tests and then write individual tests for each field we are testing, where we can modify the tested field if needed.

In any case, no strong opinions here, let's wait for @hmoniz input.

adlrocha (approved):
LGTM, lets wait for @hmoniz final input and let's get it merged 🙏

src/server/mod.rs
hmoniz: Oh, agree. The `json_rpc` path can be hardcoded. Sorry for the misunderstanding.

src/config/mod.rs
hmoniz: Let's go with @adlrocha's suggestion here. I think it strikes the right balance.

src/server/mod.rs
cryptoAtwill: nice nice! I have updated accordingly already, resolving this.

cryptoAtwill:
@hmoniz @adlrocha Just a final update for this PR before closing. For the tests, I shifted the variables around within the test mode and settle down to something like the following:
```
        let config_str = formatdoc!(
            r#"
            [server]
            ...

            [subnets]
            ...
        "#
        );

        let config = Config::from_str(config_str.as_str()).unwrap();

        check_server_config(&config.server);
        check_subnets_config(&config.subnets);
```
Should strike the balance mentioned above.

src/config/subnet.rs
```diff
@@ -1,10 +1,13 @@
+use std::collections::HashMap;
 use ipc_sdk::subnet_id::SubnetID;
 use url::Url;
 use fvm_shared::address::Address;
 use serde::Deserialize;
 
 use crate::config::deserialize::{deserialize_accounts, deserialize_subnet_id};
 
+pub type Subnets = HashMap<String, Subnet>;
```
hmoniz, line 9: nit: I know you already merged this, but I feel like defining a type for this obscures the code a bit as it adds a layer of opacity to the underlying type.

PR #23, opened 2023-02-21
Daemon cli

cryptoAtwill:
This PR adds the `daemon` CLI to launch the ipc agent json rpc server. It takes the config file path as command line input. Sample config file is placed in `examples` folder.

# Launch
```
cargo build --release
ipc-client daemon ./examples/sample_config.toml
```
@adlrocha While this PR is being reviewed, I will create another one that handles the versioning of the command line from git commit, to keep the PR small.


src/cli/commands/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+
+mod server;
+
+use crate::cli::CommandLineHandler;
+use clap::{Parser, Subcommand};
+use std::fmt::Debug;
+
+/// The `cli` method exposed to handle all the cli commands, ideally from main.
+///
+/// Consider using macro_rules! to handle the registration of new CLI handlers.
```
adlrocha, line 11: This is no longer the case, right? We don't use macros.

src/cli/commands/server.rs
```diff
@@ -0,0 +1,38 @@
...
+impl CommandLineHandler for LaunchJsonRPC {
+    type Arguments = LaunchJsonRPCArgs;
+
+    async fn handle(arguments: &Self::Arguments) -> anyhow::Result<()> {
+        env_logger::init_from_env(env_logger::Env::new().default_filter_or("info"));
+
+        log::debug!("launching json rpc server with args: {:?}", arguments);
+
+        let config = Config::from_file(&arguments.config_file)?.server;
+        log::info!("starting json rpc server at: {:}", config.json_rpc_address);
```
adlrocha, line 24:
```suggestion
        log::info!("starting IPC-agent daemon at: {:}", config.json_rpc_address);
```

src/cli/commands/server.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use crate::cli::CommandLineHandler;
+use crate::config::Config;
+use crate::server::jsonrpc::JsonRPCServer;
+
+/// The command to start the ipc agent json rpc server in the foreground.
+pub(crate) struct LaunchJsonRPC;
+
+#[async_trait]
+impl CommandLineHandler for LaunchJsonRPC {
```
adlrocha, line 15: Let's call this `LaunchDaemon` or something more generic, in the future we may launch with this CLI more processes, not only the jsonrpc server
```suggestion
impl CommandLineHandler for LaunchDaemon {
```

src/cli/commands/server.rs
```diff
@@ -0,0 +1,38 @@
...
+use async_trait::async_trait;
+use clap::Args;
+use std::fmt::Debug;
+
+use crate::cli::CommandLineHandler;
+use crate::config::Config;
+use crate::server::jsonrpc::JsonRPCServer;
+
+/// The command to start the ipc agent json rpc server in the foreground.
+pub(crate) struct LaunchJsonRPC;
```
adlrocha, line 12:
```suggestion
pub(crate) struct LaunchDaemon;
```

src/cli/commands/server.rs
```diff
@@ -0,0 +1,38 @@
+//! The Daemon command line handler that prints the info about IPC Agent.
```
adlrocha, line 1: Call this file daemon.rs better

src/cli/commands/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+            e
+        )
+    }
+}
+
+/// The collection of all subcommands to be called, see clap's documentation for usage. Internal
+/// to the current mode. Register a new command accordingly.
+#[derive(Debug, Subcommand)]
+enum Commands {
+    /// Launch the ipc agent json rpc server.
```
adlrocha, line 60:
```suggestion
    /// Launch the ipc agent daemon.
```

src/cli/commands/server.rs
```diff
@@ -0,0 +1,38 @@
...
+
+        let server = JsonRPCServer::new(config);
+        server.run().await;
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Args)]
+#[command(about = "Prints info of the the ipc agent")]
```
adlrocha, line 34: This comment is wrong, right? This should be fixed.

examples/sample_config.toml
```diff
@@ -0,0 +1,16 @@
+[server]
```
adlrocha, line 1: Instead of making this a sample_config, what about making it a `template.toml` in the `config/` for now. I am generally not a fan of this kind of sample files :)

src/cli/commands/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+/// enum Commands {
+///     NewCommand(<NewCommand as CommandLineHandler>::Arguments),
+/// }
+/// ```
+pub async fn cli() {
+    // parse the arguments
+    let args = IPCAgentCliCommands::parse();
+
+    let r = match &args.command {
+        Commands::Daemon(args) => <server::LaunchJsonRPC as CommandLineHandler>::handle(args).await,
```
adlrocha, line 44: I think I asked you at some point, but is there an alternative to make this a bit less verbose?

adlrocha (changes requested):
First quick round of suggestions. Let me know your thoughts. Also, I see in the description that you were going to handle the version issue in a follow-up PR, let me know if this is still the case (it is not on the critical path so feel free to track it on an issue and focus on the subnet manager for now).

src/cli/commands/mod.rs
cryptoAtwill: removed, resolving this

src/cli/commands/server.rs
cryptoAtwill: updated, resolving this

src/cli/commands/server.rs
cryptoAtwill: updated, resolving this

src/cli/commands/server.rs
cryptoAtwill: updated

src/cli/commands/server.rs
cryptoAtwill: udpated

examples/sample_config.toml
cryptoAtwill: updated

src/cli/commands/mod.rs
cryptoAtwill: updated, simplified the notations.

cryptoAtwill:
@adlrocha I have updated the code based on the review comments, feel free to take a look. As for the injection of version, I have added the issue: #26 to track it.

adlrocha (approved):
LGTM. Thanks! I added one last nit but it shouldn't be a blocker. Let's see if @hmoniz wants to have a quick look before merging. We should at some point think about how we are going to add tests to cli commands.

src/cli/commands/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/// The overall command line struct to be used by `clap`.
+#[derive(Debug, Parser)]
+#[command(name = "ipc", about = "The IPC agent command line tool")]
+#[command(propagate_version = true)]
+struct IPCAgentCliCommands {
+    #[command(subcommand)]
+    command: Commands,
+}
```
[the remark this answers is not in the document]
cryptoAtwill: udpated and resolving this.

hmoniz (approved):

PR #25, opened 2023-02-21
subnet manager interface and lotus impl placeholder

adlrocha:
This PR introduces a `SubnetManager` to be implemented by different subnet implementations to handle the core subnet interactions. The PR also adds a placeholder for the Lotus-specific implementation so we can start dividing the work between @cryptoAtwill and I, and we can start implementing each of the operations.


src/manager/lotus_manager.rs
```diff
@@ -0,0 +1,48 @@
...
+use anyhow::Result;
+use async_trait::async_trait;
+use fvm_shared::{address::Address, econ::TokenAmount};
+use ipc_gateway::Checkpoint;
+use ipc_sdk::subnet_id::SubnetID;
+use ipc_subnet_actor::{ConstructParams, JoinParams};
+
+use super::{SubnetInfo, SubnetManager};
+
+pub struct LotusManager {}
```
hmoniz, line 12: Consider renaming it to `LotusSubnetManager`. Even if we ultimately want a monolithic `LotusManager`, we can construct it by the composition of the submanagers.

hmoniz: Perhaps also rename this file just to `lotus.rs`.

src/manager/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+///! IPC node-specific traits.
+use anyhow::Result;
+use async_trait::async_trait;
+use fvm_shared::{address::Address, econ::TokenAmount};
+use ipc_gateway::Checkpoint;
+use ipc_sdk::subnet_id::SubnetID;
+use ipc_subnet_actor::{ConstructParams, JoinParams, Status};
+use serde::{Deserialize, Serialize};
+
+/// Trait to interact with a subnet and handle its lifecycle.
```
hmoniz, line 14: nit: Move this stuff to a `subnet` module nested in the `manager` module.

hmoniz (approved):
LGTM. Just some nits regarding code structure.

cryptoAtwill (approved):

src/manager/lotus_manager.rs
adlrocha: Good points. Fixing it.

src/manager/mod.rs
adlrocha: You mean moving the `SubnetManager` to a `manager/subnet.rs` file?

src/manager/subnet.rs
```diff
@@ -0,0 +1,57 @@
...
+use serde::{Deserialize, Serialize};
+
+/// Trait to interact with a subnet and handle its lifecycle.
+#[async_trait]
+pub trait SubnetManager {
+    /// Deploys a new subnet actor on the `parent` subnet given as an input and with the
+    /// configuration passed in `ConstructParams`.
+    /// The result of the function is the ID address for the subnet actor from which the final
+    /// subet ID can be inferred.
+    fn create_subnet(parent: SubnetID, from: Address, params: ConstructParams) -> Result<Address>;
```
cryptoAtwill, line 19: I think the ConstructParams already contains `parent`, seems the field is not needed?

cryptoAtwill, line 19: This trait will need `lotus` api to interact with the node, it needs to hold `&self`, or we are initializing the lotus api every time in the call?

adlrocha:
⏳ This pull request is set to merge as part of a Graphite merge job
Job ID: MOBemDur90CBweXhoVwt.
See details on [graphite.dev](https://app.graphite.dev/github/pr/consensus-shipyard/ipc-agent/25)

PR #27, opened 2023-02-22
add lotus rpc call

cryptoAtwill:
Add 3 json rpc calls to lotus client:

- state_network_name
- state_network_version 
- state_actor_code_cids

to assist create subnet actors in #28.

### DIFF
--- a/src/jsonrpc/tests.rs
+++ b/src/jsonrpc/tests.rs
@@ -6,7 +6,7 @@ use crate::jsonrpc::{JsonRpcClient, JsonRpcClientImpl, NO_PARAMS};
 
 /// The default endpoints for public lotus node. If the urls fail in running tests, need to
 /// check these endpoints again.
-pub const HTTP_ENDPOINT: &str = "https://api.node.glif.io/rpc/v0";
+const HTTP_ENDPOINT: &str = "https://api.node.glif.io/rpc/v0";
 const WS_ENDPOINT: &str = "wss://wss.node.glif.io/apigw/lotus/rpc/v0";
 
 #[tokio::test]
--- a/src/jsonrpc/tests.rs
+++ b/src/jsonrpc/tests.rs
@@ -4,7 +4,9 @@ use url::Url;
 
 use crate::jsonrpc::{JsonRpcClient, JsonRpcClientImpl, NO_PARAMS};
 
-const HTTP_ENDPOINT: &str = "https://api.node.glif.io/rpc/v0";
+/// The default endpoints for public lotus node. If the urls fail in running tests, need to
+/// check these endpoints again.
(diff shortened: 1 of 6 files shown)


src/lotus/tests.rs
```diff
@@ -0,0 +1,39 @@
+use url::Url;
+use crate::jsonrpc::JsonRpcClientImpl;
+use crate::lotus::{LotusClient, LotusJsonRPCClient};
+
+const HTTP_ENDPOINT: &str = "https://api.node.glif.io/rpc/v0";
```
adlrocha, line 5: Let's hope that the endpoint doesn't change :) I don't have a better option without mocking the RPC API, so I am fine with this, but maybe let's add a note mentioning that if all tests fail the RPC API may no longer be available (which may be a common issue).

src/lotus/tests.rs
```diff
@@ -0,0 +1,39 @@
+use url::Url;
```
adlrocha, line 1: We should maybe add an issue to include a test also for the rest of the methods in this API. WDYT?

adlrocha (approved):
LGTM. Thanks!

src/lotus/tests.rs
cryptoAtwill: sounds good! I will add an issue to track this.

src/lotus/tests.rs
cryptoAtwill: Added comment for this. This parameter is also used in `jsonrpc` mod, I updated the code to use import instead, so that we only have one place to change.

hmoniz: Yep. This is not ideal, but I think it's the best we can do for now. One option in the short-term, given the potential flakiness of these tests, is to mark them with `#[ignore]` and then explicitly run them with `cargo test -- --ignored`. But I think we can keep them as is for now.

src/jsonrpc/tests.rs
```diff
@@ -4,7 +4,9 @@ use url::Url;
 
 use crate::jsonrpc::{JsonRpcClient, JsonRpcClientImpl, NO_PARAMS};
 
-const HTTP_ENDPOINT: &str = "https://api.node.glif.io/rpc/v0";
+/// The default endpoints for public lotus node. If the urls fail in running tests, need to
+/// check these endpoints again.
+pub const HTTP_ENDPOINT: &str = "https://api.node.glif.io/rpc/v0";
```
hmoniz, line 9: I think I'd rather keep this private and explicitly add another constant in the other `tests` module, but up to you. It just feels weird not having each test module hermetic.

hmoniz (approved):

src/jsonrpc/tests.rs
cryptoAtwill: yeah, make sense. I'll revert to each module have their own parameter. Resolving this.

PR #28, opened 2023-02-22
Create subnet actor

cryptoAtwill:
This PR implements the create subnet actor in the subnet manager interface. It is almost a direct port of the golang implementation, except that it checks the parent instead of overwrite the parent.

In this PR, it also needs a few util function from `lotus` client, please help check out the [lotus client PR](https://github.com/consensus-shipyard/ipc-agent/pull/27) and also moving the main to lib for integration testing, #30.


src/lotus/message.rs
```diff
@@ -72,11 +71,10 @@ pub struct ReadStateResponse<State> {
 #[serde(rename_all = "PascalCase")]
 pub struct Receipt {
     #[allow(dead_code)]
-    exit_code: u32,
+    pub exit_code: u32,
+    pub r#return: String,
```
adlrocha, line 75: I know that this is not your idea/fault and it comes from Lotus, but man this is horrible haha

src/manager/lotus.rs
```diff
@@ -1,24 +1,59 @@
...
-        &self,
-        _from: Address,
-        _params: ConstructParams,
-    ) -> Result<Address> {
-        panic!("not implemented")
+impl<T: JsonRpcClient + Send + Sync> SubnetManager for LotusSubnetManager<T> {
+    async fn create_subnet(&self, from: Address, params: ConstructParams) -> Result<Address> {
+        let network_name = self.lotus_client.state_network_name().await?;
+        if params.parent.to_string() != network_name {
+            return Err(anyhow!("parent network name not match"));
```
adlrocha, line 26:
```suggestion
            return Err(anyhow!("subnet actor being deployed in the wrong parent network, parent network names do not match"));
```

src/manager/lotus.rs
```diff
@@ -1,24 +1,59 @@
...
+        let nonce = mem_push_response.nonce;
+        log::debug!(
+            "create subnet message published with cid: {message_cid:?} and nonce: {nonce:?}"
+        );
+
+        let state_wait_response = self.lotus_client.state_wait_msg(message_cid, nonce).await?;
+        let address_raw = state_wait_response.receipt.r#return;
+        log::info!("created subnet at address: {address_raw:}");
+
+        Ok(Address::from_str(&address_raw)?)
```
adlrocha, line 56: Have you tested this? Are we sure this type is returned in a string form? It would make sense as it is the type returned by the RPC call, but I would double-check it just in case.

src/manager/lotus.rs
```diff
@@ -1,24 +1,59 @@
...
+        );
+
+        let mem_push_response = self.lotus_client.mpool_push_message(message).await?;
+        let message_cid = mem_push_response.cid()?;
+        let nonce = mem_push_response.nonce;
+        log::debug!(
+            "create subnet message published with cid: {message_cid:?} and nonce: {nonce:?}"
+        );
+
+        let state_wait_response = self.lotus_client.state_wait_msg(message_cid, nonce).await?;
```
adlrocha, line 52: Now I am wondering something, and I would love your and @hmoniz input here. We are making this call synchronous, meaning that the call will wait for the message to go through and be verified to get the result. Should we add a `sync: bool` parameter or a flag when initializing the subnet manager to determine if the calls should be sync or async, and provide async versions of these methods (that don't wait for the message and the result needs to be proactively queried)? (This applies to all methods in the manager).

src/manager/lotus.rs
```diff
@@ -39,11 +74,33 @@ impl SubnetManager for LotusSubnetManager {
...
     }
 }
+
+impl<T: JsonRpcClient + Send + Sync> LotusSubnetManager<T> {
+    async fn get_subnet_actor_code_cid(&self) -> Result<Cid> {
+        let network_version = self.lotus_client.state_network_version(vec![]).await?;
+        self.state_actor_code_cids(network_version).await
+    }
+
+    async fn state_actor_code_cids(&self, network_version: NetworkVersion) -> Result<Cid> {
```
adlrocha, line 97: You are returning the code cid for the subnet actor gere, not the whole map. I would suggest renaming the function and adding a description in a comment. 
```suggestion
    async fn subnet_actor_code_cid(&self, network_version: NetworkVersion) -> Result<Cid> {
```

src/manager/params.rs
```diff
@@ -0,0 +1,14 @@
...
+
+/// Init actor Exec Params, see https://github.com/filecoin-project/builtin-actors/blob/master/actors/init/src/types.rs#L17
+#[derive(Serialize_tuple, Deserialize_tuple, Debug)]
+pub(crate) struct ExecParams {
+    pub code_cid: Cid,
+    pub constructor_params: RawBytes,
+}
+
+/// Init actor exec method number, see https://github.com/filecoin-project/builtin-actors/blob/fb759f87fcd5de0a98cb61966cd27f680df83364/actors/init/src/lib.rs#L32
+pub(crate) const INIT_EXEC_METHOD_NUM: MethodNum = 2;
```
adlrocha, line 14: Let's make this available in the `fvm_utils` instead of here. We are keeping all of these builtin_actors and fvm-specific things in that repo. This crate is a good candidate: https://github.com/consensus-shipyard/fvm-utils/tree/main/runtime/src/builtin

You could also pick it up from the `builtin_actors` but that may lead to additional overheads to comply with their version updates.

Cargo.toml
```diff
@@ -29,17 +29,20 @@ warp = "0.3.3"
...
 
 [workspace.dependencies]
+fvm_ipld_encoding = "0.3.3"
 fvm_shared = { version = "=3.0.0-alpha.17", default-features = false }
 fil_actors_runtime = { git = "https://github.com/consensus-shipyard/fvm-utils", features = ["fil-actor"] }
 ipc-sdk = { git = "https://github.com/consensus-shipyard/ipc-actors.git" }
-ipc-subnet-actor = { git = "https://github.com/consensus-shipyard/ipc-actors.git", features = [] }
-ipc-gateway = { git = "https://github.com/consensus-shipyard/ipc-actors.git", features = [] }
+ipc-subnet-actor = { git = "https://github.com/consensus-shipyard/ipc-actors.git" , features = [] }
+ipc-gateway = { git = "https://github.com/consensus-shipyard/ipc-actors.git" , features = [] }
```
adlrocha, line 47:
```suggestion
ipc-subnet-actor = { git = "https://github.com/consensus-shipyard/ipc-actors.git", features = [] }
ipc-gateway = { git = "https://github.com/consensus-shipyard/ipc-actors.git", features = [] }
```

adlrocha (changes requested):
Great job! This one shouldn't require many RTTs. Let's be sure that it works before merging testing against our version of Lotus.

src/manager/params.rs
cryptoAtwill: right right, this parameter is not used, I forgot to delete.

src/lotus/message.rs
cryptoAtwill: You mean the `return` right? We can rename this to `return_string` if we want to avoid `r#`.

src/manager/lotus.rs
cryptoAtwill: So once the message is pushed to mempool, we return the message cid and that's it. Leave it to the caller to do the follow up state wait? I think better create another method instead of using a bool if we want to support this use case. One reason is the return result would be different. With the bool, it can return Cid or Address, which we have to wrap the result in another layer. Having another method is just easier.

src/manager/lotus.rs
cryptoAtwill: Updated and resolving this.

src/manager/params.rs
cryptoAtwill: Then also `ExecParams` params too?

src/manager/lotus.rs
adlrocha: Tracked here: https://github.com/consensus-shipyard/ipc-agent/issues/31

src/manager/params.rs
```diff
@@ -0,0 +1,14 @@
+use cid::Cid;
+use fvm_ipld_encoding::RawBytes;
+use fvm_shared::MethodNum;
+use fvm_ipld_encoding::tuple::{Deserialize_tuple, Serialize_tuple};
+
+/// Init actor Exec Params, see https://github.com/filecoin-project/builtin-actors/blob/master/actors/init/src/types.rs#L17
+#[derive(Serialize_tuple, Deserialize_tuple, Debug)]
```
adlrocha, line 7: This hasn't been moved yet to fvm-utils right?

tests/integration_test.rs
```diff
@@ -0,0 +1,35 @@
...
+use ipc_sdk::subnet_id::{ROOTNET_ID};
+use ipc_subnet_actor::{ConsensusType, ConstructParams};
+use ipc_client::lotus::LotusClient;
+use ipc_client::manager::{LotusSubnetManager, SubnetManager};
+use crate::setup::{LOCAL_JSON_RPC_HTTP_URL};
+
+mod setup;
+
+#[tokio::test]
+#[ignore] // ignore this test for now as it's still developing
```
hmoniz, line 11: Maybe keep it out of the repository for now then?

src/manager/params.rs
hmoniz: +1

src/main.rs
```diff
@@ -1,8 +1 @@
-mod cli;
-mod config;
-mod jsonrpc;
-mod lotus;
-mod manager;
-mod server;
-
-fn main() {}
+fn main() {}
```
hmoniz, line 1: Maybe revert changes like these to avoid noise in the PR and in the git history.

src/manager/lotus.rs
hmoniz: Yep, my suggestion is to leave this as is for now. Later, we can create a new function that wraps this call with a verification logic that checks when the message is applied to the chain state.

hmoniz (commented):
Approving, assuming that you tested this end-to-end. If not, please make sure it works.

hmoniz (approved):

src/manager/lotus.rs
adlrocha: Resolving

adlrocha (approved):
The init-actor-related issue and Henrique's comment and we should be good to merge. Thanks :pray:

src/lotus/message.rs
cryptoAtwill: yeah, I changed to `result` instead of `return`, maybe it's better than `return`.

src/main.rs
cryptoAtwill: updated this to resolve the conflict, resolving

tests/integration_test.rs
cryptoAtwill: This integration test is working now and it won't be triggered only if explicitly run it. Probably we can keep it for now to be example and use it as some sanity test for checking any updates to `lotus` that might break rust side. Eventually it will be moved to json rpc server, once this happens, I will delete this file. Hope this is ok.

src/manager/params.rs
cryptoAtwill: moved, resolving this

README.md
```diff
@@ -1,2 +1,7 @@
 # ipc-agent
 Implementation of an IPC agent
+
+## Dev
```
adlrocha, line 4: Can we move this message somewhere else? Maybe into a `notes.md` in the root of the repo? I don't think it fits the main README.

Also, would you mind adding a bit more of context of where and how should the address be set for the first bullet point?

src/lotus/message.rs
adlrocha: In the end we needed a renaming. Resolving this.

tests/integration_test.rs
adlrocha: What about moving this to a gits file outside this repo and pointing to it from the dev notes you started above? I am just worried that we may forget to remove it and updated in the future leading to dirt in the repo.

adlrocha (approved):
A few last-minute comments after our tests from this morning. But feel free to merge after addressing them. Thanks!

PR #30, opened 2023-02-23
move to lib

cryptoAtwill:
To migrate `main.rs` to use both `main.rs` and `lib.rs`, reasons are as following:
1. Currently we are using `main.rs`, but it's empty and it leads to a lot of clippy dead code warning. Moving to `lib.rs` will resolve these warnings.
2. When writing integration tests for `SubnetManager`, it needs to import modules from `ipc-agent`, hence it's only possible if we change to lib.
3. Seems doc tests are compiled only with lib, see https://github.com/rust-lang/rust/issues/50784, I could be wrong, but somehow `cargo test` does not trigger the doc tests, with `lib` it's just easy.

Since there is the chance, I also fixed a few clippy warning, so that we are now warning free!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,9 +33,13 @@ thiserror = "1.0.38"
 fvm_shared = {workspace = true}
 fil_actors_runtime = {workspace = true}
 ipc-sdk = { workspace = true }
+ipc-subnet-actor = { workspace = true }
+ipc-gateway = { workspace = true }
 
 [workspace.dependencies]
 fvm_shared = { version = "=3.0.0-alpha.17", default-features = false }
 fil_actors_runtime = { git = "https://github.com/consensus-shipyard/fvm-utils", features = ["fil-actor"] }
 ipc-sdk = { git = "https://github.com/consensus-shipyard/ipc-actors.git" }
+ipc-subnet-actor = { git = "https://github.com/consensus-shipyard/ipc-actors.git", features = [] }
+ipc-gateway = { git = "https://github.com/consensus-shipyard/ipc-actors.git", features = [] }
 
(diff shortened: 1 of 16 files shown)


Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "ipc-client"
+name = "ipc-agent"
```
adlrocha, line 2: Good catch :)

adlrocha (approved):
LGTM. Let's get a 👍 from @hmoniz that he is OK with this approach too and let's get it merged.

src/config/mod.rs
```diff
@@ -21,29 +21,28 @@ pub const JSON_RPC_VERSION: &str = "2.0";
...
+pub struct Config {
     pub server: Server,
     pub subnets: HashMap<String, Subnet>,
 }
 
 impl Config {
     /// Reads a TOML configuration in the `s` string and returns a [`Config`] struct.
-    pub fn from_str(s: &str) -> Result<Self> {
-        let config = toml::from_str(&s)?;
+    pub fn from_string(s: &str) -> Result<Self> {
```
aakoshh, line 31: Just curious, why this rename? If it's `from_str`, then why not implement [FromStr](https://doc.rust-lang.org/std/str/trait.FromStr.html), and if that doesn't feel right, then why not call it `from_toml` or `from_toml_str`? I suppose mentioning `toml` in the method name would also put `from_file` into question.

hmoniz (approved):
LGTM. Assuming you'll address Akosh's comments.

Cargo.toml
aakoshh: I named my packages with underscores following the convention in `ref-fvm`, and the `builtin-actors`, although apparently not in `rust-libp2p`. I thought it might convert them to underscores, but no, when using it in cargo it would be eg. `cargo test -p ipc-agent`, and `cargo test -p ipc_ipld_resolver`. 

Maybe we could all stick to one?

PR #41, opened 2023-02-24
Add a method to LotusClient for retrieving the chain head

hmoniz:
The first of a series of changes to add new JSON-RPC method calls to the `LotusClient` necessary for checkpointing implementation. This one adds the `ChainHead` call. The deserialization is done into a simplified data structure that does not decode the entire response. In particular, we leave out the `blocks` field due to its complexity which is unnecessary for our purposes.


adlrocha (approved):
LGTM. Something I've been wondering for the checkpoint process, will you use polling and http for the lotus client, or will you need to subscribe to the web socket?

hmoniz:
> LGTM. Something I've been wondering for the checkpoint process, will you use polling and http for the lotus client, or will you need to subscribe to the web socket?

I prefer HTTP polling simply because I don't trust WS reliability enough. It's a discussion we have have though.

PR #42, opened 2023-02-24
Splits lotus/message module into submodules

hmoniz:
Some housekeeping. This PR splits the `message` module into submodules, one for each category of calls.


src/lotus/message/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+pub(crate) mod state;
+pub(crate) mod wallet;
+
+/// Exec actor parameters
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "PascalCase")]
+pub struct ExecParams {
+    pub code_cid: Cid,
+    pub constructor_params: Vec<u8>,
+}
```
adlrocha, line 20: @cryptoAtwill I think this are now part of the fvm_utils and can be removed from the agent, right? It may have been left behind from your PR.

src/lotus/message/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+
+/// Install actor parameters
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "PascalCase")]
+pub struct InstallActorParams {
```
adlrocha, line 25: @cryptoAtwill, same here, do we still need this?

adlrocha (approved):
LGTM. As @cryptoAtwill may be asleep, feel free to remove the types I tagged if you see that nothing breaks. If something breaks I can look at it or wait for him, this morning we merged a PR that moves those types to fvm_utils. 

Also, what do you think about renaming the package to `types` instead of `messages`? They seem more of lotus types than messages.

src/lotus/message/mod.rs
hmoniz: I'll remove this

hmoniz:
> LGTM. As @cryptoAtwill may be asleep, feel free to remove the types I tagged if you see that nothing breaks. If something breaks I can look at it or wait for him, this morning we merged a PR that moves those types to fvm_utils.
> 
> Also, what do you think about renaming the package to `types` instead of `messages`? They seem more of lotus types than messages.

This leaves me very conflicted. I dislike the `types` pattern because, over time, it tends to become a dumping ground for random type definitions leading to a loss of structure. I wanted this module to contain the types that represent the data exchanged during JSON-RPC calls. To that effect, `message` seems a more specific name than the generic `types`. If you strongly prefer `types`, however, I can rename it. Wdyt?

src/lotus/message/mod.rs
cryptoAtwill: It can be removed as well.

src/lotus/message/mod.rs
cryptoAtwill: thx thx!

cryptoAtwill (approved):

PR #46, opened 2023-02-28
create subnet

cryptoAtwill:
Add `create_subnet` method handler to the json rpc server.

The implementation of the `create_subnet` logic is moved to another PR as it can get quite involved. Just the skeleton to interact with the server first.


src/server/handlers/create.rs
```diff
@@ -0,0 +1,35 @@
...
+/// The create subnet json rpc method handler.
+pub(crate) struct CreateSubnetHandler {}
+
+#[async_trait]
+impl JsonRPCRequestHandler for CreateSubnetHandler {
+    type Request = CreateSubnetParams;
+    type Response = CreateSubnetResponse;
+
+    async fn handle(&self, _request: Self::Request) -> anyhow::Result<Self::Response> {
+        Ok(CreateSubnetResponse{ address: String::from("/root") })
```
hmoniz, line 33: I guess this is just a placeholder, right?

PR #48, opened 2023-02-28
Adds IPC-related JSON-RPC methods for the LotusClient

hmoniz:
This PR adds the `IPCGetCheckpointTemplate` and the `IPCGetCheckpointForPrevChild` methods to the `LotusClient`. Tests are coming in a subsequent PR as, for those, we need to set up subnets with the subnet manager.


src/lotus/client.rs
```diff
@@ -212,4 +221,35 @@ impl<T: JsonRpcClient + Send + Sync> LotusClient for LotusJsonRPCClient<T> {
...
+        child_subnet_id: SubnetID,
+    ) -> Result<IPCGetPrevCheckpointForChildResponse> {
+        let parent = match child_subnet_id.parent() {
+            None => return Err(anyhow!("The child_subnet_id must be a valid child subnet")),
+            Some(parent) => parent,
+        };
+        let subnet_actor = child_subnet_id.subnet_actor().to_string();
+        let params =
+            json!([GATEWAY_ACTOR_ADDRESS, {"Parent": parent.to_string(), "Actor": subnet_actor}]);
+        println!("{}", params);
```
cryptoAtwill, line 236: probably a missed `println` for log::info?

cryptoAtwill (approved):

src/lotus/client.rs
hmoniz: oops

PR #49, opened 2023-02-28
IPC-35: Config

### LINKED ISSUE #35: IPLD Resolver: Membership
Part of #17 

We want to know which IPC agents we can contact to resolve content from a specific subnet. To do so, we can use `Gossipsub`. 

First, I'm assuming we have a single P2P network of agents, with everyone potentially able to connect to anyone else, with every agent running a single `Swarm` on a single address, serving the needs of all subnets. Later we can consider alternatives.

There are at least two ways to use `Gossipsub` to achieve what we want:
* Agents who are able to serve data from a subnet can _publish_ a message into a topic, e.g. `/ipc/subnet-agents`. The message will spread across the network, and every node can build up a view of which agents do which subnets (a bipartite graph, as agents can participate in multiple subnets). 
* We can potentially forego even publishing if we use the topics themselves to signal membership. Agents who are able to serve data from a subnet can _subscribe_ to a topic, e.g. `/ipc/subnet/<subnet-id>`. The subscription information spreads around the network and at least the Rust version of `Gossipsub` can be queried to list peers subscribed to any topic. The parent subnet agents would not subscribe to the child subnets, they would just query their `Gossipsub` behaviour who are the subscribers of a child subnet, and contact them when they need to. The drawback here is that we only learn the subscriptions of peers [we are connected to](https://github.com/libp2p/rust-libp2p/blob/v0.50.0/protocols/gossipsub/src/behaviour.rs#L3115), wheras with the first approach we learn about everyone.

Through `Gossipsub`, we would learn about `PeerId`s that are serving subnets, but not their address. To connect to them, we rely on #34 to learn their address. 

In this PR, we develop a `Membership` behaviour, such that:
* It wraps a `Gossipsub` instance 
* It has a `join` and a `leave` method that the agent can call to subscribe or publish to the necessary topics to signal membership to others
* It listens to events from `Gossipsub` and either maintains a list of peer-subnet mappings, or is able to ask `Gossipsub` for the information (if it has to maintain it, we need to think about limits on the size of the collection).
* (Optional) It raises events when it learns about new members in a topic, which could be used to instruct the `Discovery` behaviour to ensure we proactively learn their addresses, unless they are already known, by doing Kademlia lookups.
* It provides a `members` method to return the list of peer IDs of data providers in a subnet.
* If we publish, then periodically re-publish the current subnet membership information about the agent itself, to make sure new joiners are informed. (We could publish a list, which would mean we don't even need `join`/`leave`, just a `serve` message with all the subnets, and a timestamp).
* If we publish, then track the last time we heard about the membership of an agent, and stop suggesting connecting to it if it's been too long - this is like a heartbeat so we don't end up trying to connect to agents long gone.

Questions:
* How does a new joiner learn about the current memberships? Does Gossipsub receive subscriptions from peers it connects to? Should we periodically publish our memberships to keep them fresh in everyone's tables?
* How does Gossipsub decide who to connect to in the first place? There is an option for explicit peers (we could use the same list as we do for bootstrapping Kademlia, but it looks like it also tries to connect to anyone whenever a connection is established by the Swarm, so the natural workings of Kademlia queries will prompt Gossipsub to connect as well.

Overall, the explicit publishing of messages is probably better, because: 
* The Go implementation would not serve subscription lists, it would only work for Rust. 
* It allows us to learn about peers we aren't connected to at the moment, but we might want to if we need their data, as opposed to only knowing about, say, the 50 agents we are actually connected to.

aakoshh:
Closes #35 

Follow up to #44, this PR adds configuration to the `membership` module and refactors the `discovery` module to get rid of the `ConfigBuilder` and extract common configuration fields into a `NetworkConfig`. 

It also adds `pin_subnet` methods to the `Behaviour` and the `SubnetProviderCache`, which can be used to let the module know about the existence of some known subnet IDs that it should never prune from the cache. This is to prevent attackers trying to throw tons of random subnets at us and prevent us from overcoming their numbers and learn who to resolve the content of real subnets from.

The reason I removed the builder is because I found having to repeat fields 3 times (`Config`, `ConfigBuilder` and `Behaviour`) annoying, and in the end it still wasn't enough to nicely prevent the `Behaviour::new` from having to return a `Result`, or else do `expect` and potentially panic. Since this is meant to be a library, that's not nice. It's easier to just embrace the fact that `new` does validation. We can still do our best to validate the configuration after we read it from files.

### DIFF
--- a/ipld/resolver/src/behaviour/discovery.rs
+++ b/ipld/resolver/src/behaviour/discovery.rs
@@ -10,7 +10,7 @@ use std::{
 };
 
 use libp2p::{
-    core::{connection::ConnectionId, identity::PublicKey},
+    core::connection::ConnectionId,
     kad::{
         handler::KademliaHandlerProto, store::MemoryStore, Kademlia, KademliaConfig, KademliaEvent,
         QueryId,
@@ -26,6 +26,8 @@ use libp2p::{
 };
 use log::debug;
 use tokio::time::Interval;
+
+use super::NetworkConfig;
 
 // NOTE: The Discovery behaviour is largely based on what exists in Forest. If it ain't broken...
 // NOTE: Not sure if emitting events is going to be useful yet, but for now it's an example of having one.
@@ -50,10 +52,6 @@ pub enum Event {
 /// Configuration for [`discovery::Behaviour`].
 #[derive(Clone, Debug)]
 pub struct Config {
-    /// Our own peer ID, needed to bootstrap Kademlia.
-    local_peer_id: PeerId,
-    /// Name of the network in the Kademlia protocol.
-    network_name: String,
     /// Custom nodes which never expire, e.g. bootstrap or reserved nodes.
     ///
     /// The addresses must end with a `/p2p/<peer-id>` part.
@@ -62,23 +60,6 @@ pub struct Config {
     pub target_connections: usize,
     /// Option to disable Kademlia, for example in a fixed static network.
     pub enable_kademlia: bool,
-}
-
-impl Config {
-    /// Create a default configuration with the given public key.
-    pub fn new(local_public_key: PublicKey, network_name: String) -> Result<Self, ConfigError> {
-        if network_name.is_empty() {
-            Err(ConfigError::InvalidNetwork(network_name))
-        } else {
-            Ok(Self {
-                local_peer_id: local_public_key.to_peer_id(),
-                network_name,
-                user_defined: Vec::new(),
-                target_connections: usize::MAX,
-                enable_kademlia: true,
-            })
-        }
-    }
 }
 
 #[derive(thiserror::Error, Debug)]
@@ -113,10 +94,13 @@ pub struct Behaviour {
 
 impl Behaviour {
     /// Create a [`discovery::Behaviour`] from the configuration.
-    pub fn new(config: Config) -> Result<Self, ConfigError> {
+    pub fn new(nc: NetworkConfig, dc: Config) -> Result<Self, ConfigError> {
+        if nc.network_name.is_empty() {
+            return Err(ConfigError::InvalidNetwork(nc.network_name));
+        }
         // Parse static addresses.
         let mut user_defined = Vec::new();
-        for multiaddr in config.user_defined {
+        for multiaddr in dc.user_defined {
             let mut addr = multiaddr.clone();
             if let Some(Protocol::P2p(mh)) = addr.pop() {
                 let peer_id = PeerId::from_multihash(mh).unwrap();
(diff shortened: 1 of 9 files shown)


ipld/resolver/src/behaviour/discovery.rs
```diff
@@ -153,39 +93,54 @@ pub struct Behaviour {
...
+                static_addresses.push((peer_id, addr))
+            } else {
+                return Err(ConfigError::InvalidBootstrapAddress(multiaddr));
+            }
+        }
+
+        let kademlia_opt = if dc.enable_kademlia {
             let mut kad_config = KademliaConfig::default();
-            let protocol_name = format!("/ipc/kad/{}/kad/1.0.0", config.network_name);
+            let protocol_name = format!("/ipc/kad/{}/kad/1.0.0", nc.network_name);
```
adlrocha, line 115: What about removing a kad? :)
```suggestion
            let protocol_name = format!("/ipc/{}/kad/1.0.0", nc.network_name);
```

ipld/resolver/src/behaviour/membership.rs
```diff
@@ -45,17 +46,37 @@ pub enum Event {
...
 /// peer is able to resolve CIDs in different subnets.
 pub struct Behaviour {
     /// [`Gossipsub`] behaviour to spread the information about subnet membership.
     inner: Gossipsub,
     /// Events to return when polled.
     outbox: VecDeque<Event>,
-    /// [`Keypair`] used to construct [`SignedProviderRecord`] instances.
-    keypair: Keypair,
+    /// [`Keypair`] used to sign [`SignedProviderRecord`] instances.
+    local_key: Keypair,
```
adlrocha, line 77: This local key is the one for the PeerID right?

ipld/resolver/src/behaviour/membership.rs
```diff
@@ -70,14 +91,61 @@ pub struct Behaviour {
...
+        let mut gossipsub = Gossipsub::new(
+            MessageAuthenticity::Signed(nc.local_key.clone()),
+            gossipsub_config,
+        )
+        .map_err(|s| ConfigError::InvalidGossipsubConfig(s.into()))?;
+
+        gossipsub
+            .with_peer_score(
+                scoring::build_peer_score_params(membership_topic.clone()),
+                scoring::build_peer_score_thresholds(),
```
adlrocha, line 121: We may need to explore additional config parameters to avoid extreme load. In Go impl of Gossipsub there's the concept of `TopicValidators` that I don't know if in its Rust counterpart are needed, let's consider it for the future

ipld/resolver/src/behaviour/membership.rs
```diff
@@ -239,3 +316,22 @@ impl NetworkBehaviour for Behaviour {
         Poll::Pending
     }
 }
+
+// Forest has Filecoin specific values copied from Lotus. Not sure what values to use,
+// so I'll leave everything on default for now. Or maybe they should be left empty?
```
adlrocha, line 321: I think default is the conservative decision here, once we start testing we should immediately see if this is a good choice for as first deployment.

adlrocha (approved):
LGTM

ipld/resolver/src/behaviour/discovery.rs
aakoshh: OMG :see_no_evil: 

Comes from Forest but how can I be so blind :flushed:

ipld/resolver/src/behaviour/membership.rs
aakoshh: Definitely, the Forest stuff is [rather elaborate](https://github.com/ChainSafe/forest/blob/v0.6.0/node/forest_libp2p/src/gossip_params.rs), a copy of Lotus. 

I just had no idea how Filecoin specific it is. For example it has rates on the expected number of published messages, which is different for us, so it's obviously not something we can copy in, but the rest, I can't say.

PR #50, opened 2023-02-28
CHORE: Setup a CI workflow and add license headers

aakoshh:
The PR adds SPDX license headers to every file and a CI pipeline to enforce formatting and linter rules, then run the tests.

### `Cargo.lock` file

I checked in the `Cargo.lock` file, because the agent is not just a Rust library but a Rust application, and Rust apps can have their `Cargo.lock` files in source control. This way I can also use it for caching dependencies on CI. A library would have to use `generate-lock` to do the same thing. 

Although, I'm not sure this is the best approach actually, because a single change in dependencies invalidates the whole cache. In Forest they just use the `rust-toolchain.toml` file, which I didn't understand when I looked at it, but since then I learned that Github Actions isolates caches across branches, so once the PR is merged, `main` won't have access to the cache produced by the PR; on the other hand the PR has access to the caches produced by `main`. So maybe it would be better to use a key that stays the same, and update it when something changes, to get more reuse out of caches. On the other hand, it would result in more updates, and every update at the moment adds a timestamp, so they are all different files, around 500MB. This would quickly push out older caches if someone pushes lots of commits. Perhaps this is why in Forest they don't do updates on pull requests.

#### Many `Cargo.lock` files

I was struggling to get caching working with the `**/Cargo.lock` file expression I started out with. It turns out there are multiple ones _after_ the build:

```console
❯ find . -name 'Cargo.lock'
./Cargo.lock
./target/release/wbuild/ipc-gateway/Cargo.lock
./target/release/wbuild/ipc-subnet-actor/Cargo.lock
./target/debug/wbuild/ipc-gateway/Cargo.lock
./target/debug/wbuild/ipc-subnet-actor/Cargo.lock
```


cryptoAtwill (approved):

cryptoAtwill:
btw, @aakoshh, have you considered trying out `cargo-husky` as well. It sets up the similar pipeline in local development env. We can setup `pre-commit` or `pre-push` hooks so that we enforce the linting and testing before commit or push locally instead of waiting on cicd to tell us.

hmoniz (approved):

aakoshh:
@cryptoAtwill I vaguely remember a project using husky, but for some reason it didn't make a lasting impression. Someone was especially annoyed that a tool would "mess with their github hooks". 

Out of these two, I think `pre-commit` for linting and testing would be too aggressive; IIRC I heard that many of us on the team commit things that aren't in working order, like they wouldn't compile or pass tests. It would probably slow down the workflow as well: most of the times I only run the tests of the module I'm working on, not the whole project; that's what CI is there for :) 

`pre-push` could be a good place for linting, but even here I would be hestiant if we should include clippy, or just the pre-compile checks, as it could take rather long.

I used `pre-commit` hooks in `builtin-actors` and `ref-fvm` to refresh charts from data collected by the tests, to make sure I don't forget to check them in and don't have to do manual steps. I pasted the `pre-commit` example in the PR, but because we can have only one, I left it up to readers to decide if they want to combine it with whatever they have. 

Maybe we can also include recommended hooks in the README to start with? I mean, as long as CI won't let broken code to be merged, it's up to you if you are personally more annoyed by having to fix something because the lint failed on CI, or wait longer before you can even push something. 

On a note on `cargo-husky` itself: I see that it [generates](https://github.com/rhysd/cargo-husky#usage) a hook to run `cargo test` on `pre-commit`, but in our project layout, if I just run `cargo test` then it will not test the `ipc_ipld_resolver`, only the root `ipc_agent` for some reason; I have to pass `--workspace` for it to work. I don't know why, in Fendermint the default tests everything as far as I can see; the only difference is that one doesn't have a root package. 

So we would have to use the [user-hooks](https://github.com/rhysd/cargo-husky#user-hooks) feature and check in scripts that call `make` - hopefully these will override the default behaviour and not run an extra round of tests. Then I suppose these can even serve as examples, and if someone doesn't have `cargo-husky` installed, then nothing happens for them.

PR #52, opened 2023-03-01
Create subnet json rpc server

cryptoAtwill:
This PR adds the implementation for the create subnet in JSON RPC server so that it will interact with the lotus api.

One key choice made is created an struct named: `SubnetManagerShared` that will be shared by different subnet manager related handlers.

Since we will have `create_subnet`, `join_subnet` and etc, we don't have to initialize the different handler with their own manager instances. We just need those handlers to hold the common `SubnetManagerShared` to manage the subnet manager instances. This would help us to expanding to connection pooling if there is a need.

This PR depends on #54. Please review and merge #54 first.

### DIFF
--- a/src/server/handlers/create.rs
+++ b/src/server/handlers/create.rs
@@ -1,0 +1,35 @@
+//! Create subnet handler and parameters
+
+use async_trait::async_trait;
+use fvm_shared::clock::ChainEpoch;
+use serde::{Deserialize, Serialize};
+use crate::server::JsonRPCRequestHandler;
+
+#[derive(Debug, Deserialize)]
+pub struct CreateSubnetParams {
+    pub parent: String,
+    pub name: String,
+    pub min_validator_stake: u64,
+    pub min_validators: u64,
+    pub finality_threshold: ChainEpoch,
+    pub check_period: ChainEpoch,
+}
+
+#[derive(Debug, Serialize)]
+pub struct CreateSubnetResponse {
+    /// The address of the created subnet
+    pub address: String
+}
+
+/// The create subnet json rpc method handler.
+pub(crate) struct CreateSubnetHandler {}
+
+#[async_trait]
+impl JsonRPCRequestHandler for CreateSubnetHandler {
+    type Request = CreateSubnetParams;
+    type Response = CreateSubnetResponse;
+
+    async fn handle(&self, _request: Self::Request) -> anyhow::Result<Self::Response> {
+        Ok(CreateSubnetResponse{ address: String::from("/root") })
+    }
+}
--- a/src/server/handlers/create.rs
+++ b/src/server/handlers/create.rs
@@ -1,8 +1,18 @@
 //! Create subnet handler and parameters
 
+use std::str::FromStr;
+use std::sync::Arc;
+use anyhow::anyhow;
 use async_trait::async_trait;
 use fvm_shared::clock::ChainEpoch;
+use fvm_shared::econ::TokenAmount;
+use ipc_sdk::subnet_id::SubnetID;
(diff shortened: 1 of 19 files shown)


src/server/handlers/create.rs
```diff
@@ -22,14 +32,49 @@ pub struct CreateSubnetResponse {
...
-        Ok(CreateSubnetResponse{ address: String::from("/root") })
+    async fn handle(&self, request: Self::Request) -> anyhow::Result<Self::Response> {
+        let parent = &request.parent;
+
+        if !self.shared.contains_subnet(parent) {
+            return Err(anyhow!("target parent subnet not found"));
+        }
+
+        // this is safe to unwrap as we ensure this key exists.
+        let bundle = self.shared.get(parent).unwrap();
```
adlrocha, line 58: This is really confusing, I had to read the whole code several times to understand what was happening under the hood. I think the `SharedSubnetManager` can be significantly simplified. We can discuss sync if it is easier, but I would suggest:
- I would maybe call it `SubnetManagerPool` instead of shared.
- Having a `new_pool(subnets) that initializes the managers pool for all subnets. 
- We could have a `.get(subnetID)` method that returns the manager for the subnet from the pool (similar to how you use it).
- We should have a `pool_update` to update the manager when the config is updated. We can maybe lazily do this directly in `get`, i.e. to check if we have updates

src/server/handlers/create.rs
```diff
@@ -22,14 +32,49 @@ pub struct CreateSubnetResponse {
 }
 
 /// The create subnet json rpc method handler.
-pub(crate) struct CreateSubnetHandler {}
+pub(crate) struct CreateSubnetHandler<T: JsonRpcClient> {
```
adlrocha, line 35: We may want this to be a `SubnetHandler` that aggregates all subnet manager methods that way we only have to pass the manager pool once.

src/server/handlers/mod.rs
```diff
@@ -43,4 +56,24 @@ impl Handlers {
...
     }
+
+    /// Create the needed subnet managers for each subnet.
+    ///
+    /// Since we don't have a large number of subnet for now, to keep things simple,
+    /// these managers are created upon initialization.
+    ///
+    /// If the traffic received by the json rpc node increases or the number of subnets increases
+    /// significantly, we can use Connection Pooling for manage the subnet managers.
+    fn create_managers(subnets: &HashMap<String, Subnet>) -> HashMap<String, LotusSubnetManager<JsonRpcClientImpl>> {
```
adlrocha, line 67: I don't understand why can't you do this when you initialize your pool/shared manager, It would be  cleaner.

src/server/handlers/subnet.rs
```diff
@@ -0,0 +1,46 @@
+//! The shared subnet manager module for all subnet management related RPC method calls.
+
+use anyhow::{anyhow, Result};
+use std::collections::HashMap;
+use crate::config::Subnet;
+use crate::jsonrpc::JsonRpcClient;
+use crate::manager::LotusSubnetManager;
+
+pub(crate) struct ManagerConfigBundle<T: JsonRpcClient> {
```
adlrocha, line 9: I would embed all of this as part of the pool/manager object to avoid having it independently.

adlrocha (changes requested):
I think we can simplify this significantly as mentioned in the comments, and we should consider config updates. We can discuss sync if it is easier. Thanks 🙏

src/server/handlers/mod.rs
cryptoAtwill: because if we init in the manager, then manager will have the `JsonRPCClient` fixed. The type should be determined in the caller.

PR #54, opened 2023-03-01
Support config reloading

cryptoAtwill:
Add support to config hot reloading.

This is using the `config` crate's watch capability, see [example](https://github.com/mehcode/config-rs/blob/master/examples/watch/main.rs). 

Initially, there are two choices, hot reloading and lazy reloading. Lazy reloading is triggered if some key is not found in the config. Hot reloading is constantly watching the config file and only triggers reload when there is a change. 

After some try and error, hot reloading is preferred as it might be more DDOS friendly. The look up of keys in config will be exposed in json rpc calls, if external users sent random keys, then lazy reloading will always be triggered. Hot reloading does not have this problem as if key not found, then it's not there, it's not dependent on external calls.

Feel free to comment otherwise.

See `reload_works` test case in `tests.rs`.


src/config/hot.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub fn new_with_watcher(path: String, watch_interval: u64) -> Result<Self> {
+        let (stop_signal_tx, stop_signal_rx) = channel();
+        let (reload_notify_tx, reload_notify_rx) = channel();
+
+        let config = Arc::new(RwLock::new(ConfigWrapper {
+            config: Config::from_file(path.clone())?,
+        }));
+
+        let config_cloned = config.clone();
+        thread::spawn(move || {
```
adlrocha, line 30: Let's try if we can use a tokio thread so we use a green thread intead a physical one. Maybe the watcher doesn't support it but let's see if we can save some resources considering that we already have the tokio runtime.

src/config/tests.rs
```diff
@@ -0,0 +1,189 @@
...
+    let (tx, rx) = channel();
+    let t1 = thread::spawn(move || {
+        let mut loop_num = 1;
+
+        // we need to loop here because we need to make sure the
+        // config update is picked up by the watcher thread.
+        // It is possible that when the `file.write_all` is done,
+        // `wait_for_modification` has not yet being called. Need
+        // to loop and update to make sure the update is propagated.
+        loop {
```
adlrocha, line 74: Let's use a barrier to sync these two threads (it's hard to understand what is happening here).

adlrocha (changes requested):
Some comments, let's try the apply config approach see if is cleaner.

Issue #34, opened 2023-02-24
IPLD Resolver: Discovery


aakoshh:
Part of #17 

In order to pull content from other agents, we need to know where they are. Most of the libp2p protocols assume the existence of some "ambient" peer discovery process, meaning there is a protocol which learns about peer addresses. 

In libp2p, most of the time we refer to peers by their `PeerId` (some kind of hash of their public key), for example when we want the `Swarm` to `Dial` a peer, we emit an event with just the `PeerId`, although it is possible to attach a `MultiAddress` as well. When the `Swarm` processes this event, it calls the `NetworkBehaviour::addresses_of_peer` method to ask any of the protocols whether they know where the peer is (unless the address was given in the `Dial` options). 

In Forest, there is a `DiscoveryBehaviour` that does multiple things:
* contains a user-supplied list of hardcoded addresses
* optionally a Kademlia behaviour instance, which it calls periodically to do random lookups to find new peers
* optionally an mDNS instance, for local multicast based peer discovery

We also learned from the Bacalhau team that they rely exclusively on `Gossipsub` for peer discovery by enabling the Peer Exchange part of the protocol. Peer Exchange (a.k.a. PX) means that when a peer sends a PRUNE to another due to over subscription of a mesh (ie. it wants it to go find others to connect to), it also sends a list of alternative peers. It does so by sending Signed Address Records, which is essentially the address signed by the public key corresponding to the peer ID of the alternative party, to resist tampering by the pruner. In the Go implementation of the library this works, however in Rust the signed records are not implemented and therefore it can only connect to peers who it already knows the address for, ergo it doesn't work for peer discovery. 

In this ticket, we implement a `Discovery` behaviour for the agent:
* Start from a user-supplied list of peers, for static networks and bootstrapping.
* Use Kademlia, but make it optional, so we can disable it in small local networks if we want to.
* (Optional) Take hints from Forest about the features of the module: e.g. track the number of connected peers and pause discovering if we have enough of them, and also to slow down lookups after the bootstrap period.
* (Optional) Serve the list of currently connected peers, although this is pretty simple for any other module to track by handling Swarm events.
* (Optional) Classify peer IDs and connected/known/unknown
* Serve the list of known peer addresses from the user list and by asking Kademlia
* Start looking up specific peer IDs of subnets as we learn about them

PR #55, opened 2023-03-01
IPC-37: Resolver Service and Client

### LINKED ISSUE #37: IPLD Resolver: Service
Part of #17 

Create an IPLD Resolver Service that:
* Creates a `Swarm` with an IPLD Resolver Behaviour wrapping #34 #35 and #36
* Has a `run` method to start listening on an address and poll the `Swarm` for events
* Interprets certain events that need to go between the constituent behaviour protocols, e.g. when `Membership` raises an event about an agent serving a new subnet it didn't know about before, then the service would prompt `Discovery` to look up its peer ID to ensure it has the address if needed. 
* Has a _command_ enum to accept tasks to do over a channel from the rest of the program. Command would arrive with response channels - these would be things like resolving up CIDs. 
* Has a command queue which it also polls (along with the `Swarm`) to look for internal requests.
* Wraps the command queue into an async interface where the response channels are created and completed, for convenience. Instances of this interface can be shared out to the application and they can be cloned as well, and once the service is running, this is the only way to talk to it.
* Configuration which aggregates the configs of all constituent behaviours.
* When running the _resolve_ command, it should ask `Membership` for the list of peer IDs serving data from a topic, ask `Discovery` for the list of connected/known addresses, then decide how many to pass to `Resolve` and in which order - connected first, known addresses last; but maybe just a few at a time to not span the network, e.g. if we know 100 agents serving data in a subnet, we can send bitswap requests to 10 of them, and if it fails, then another 10, etc. Note that the `Bitswap` library is clever enough to only send `want-have` first to all-but-one, and then `want-block` to one at a time, but we might want to keep even the `want-have` within a limit. At some point we might even ask the parent subnet members for the data - a fully Bitswap implementation would keep the wants and complete them later, but not `libp2p-bitswap`.

aakoshh:
Closes #37 

The PR creates a `Service` and `Client` to act as the public interface of content resolution. 

The `Service` is what users of the library have to instantiate, it creates all the behaviours created in #34 , #35 and #36 , along with a `Client`, which can be used to send requests/commands to it over a channel. The `Client` can be cloned, so that a separate instance can be given to any component that needs to talk to the `Service`. 

Instead of an request/response queue as in Forest, the `Client` hides the query behind an async `resolve` method. The caller needs to tell which CID they want to resolve and from which subnet. The `Service` returns an error immediately if there are no peers known in that subnet.

The other two methods at the moment are to set a list of subnets the agent can provide data for, and to pin subnets which have known IDs (from config or from the ledger). Both take a list, but the first overrides any current value, the second adds to them.

### DIFF
--- a/ipld/resolver/src/behaviour/mod.rs
+++ b/ipld/resolver/src/behaviour/mod.rs
@@ -8,11 +8,16 @@ use libp2p::{
     swarm::NetworkBehaviour,
     PeerId,
 };
+use libp2p_bitswap::BitswapStore;
 
 mod content;
 mod discovery;
 mod membership;
 
+pub use discovery::Config as DiscoveryConfig;
+pub use membership::Config as MembershipConfig;
+
+#[derive(Clone, Debug)]
 pub struct NetworkConfig {
     /// Cryptographic key used to sign messages.
     pub local_key: Keypair,
@@ -29,22 +34,53 @@ impl NetworkConfig {
     }
 }
 
-/// Libp2p behaviour to manage content resolution from other subnets, using:
+#[derive(thiserror::Error, Debug)]
+pub enum ConfigError {
+    #[error("Error in the discovery configuration")]
+    Discovery(#[from] discovery::ConfigError),
+    #[error("Error in the membership configuration")]
+    Membership(#[from] membership::ConfigError),
+}
+
+/// Libp2p behaviour bundle to manage content resolution from other subnets, using:
 ///
 /// * Kademlia for peer discovery
 /// * Gossipsub to advertise subnet membership
 /// * Bitswap to resolve CIDs
 #[derive(NetworkBehaviour)]
-pub struct IpldResolver<P: StoreParams> {
+pub struct Behaviour<P: StoreParams> {
     ping: ping::Behaviour,
     identify: identify::Behaviour,
     discovery: discovery::Behaviour,
     membership: membership::Behaviour,
-    bitswap: content::Behaviour<P>,
+    content: content::Behaviour<P>,
 }
 
 // Unfortunately by using `#[derive(NetworkBehaviour)]` we cannot easily inspects events
 // from the inner behaviours, e.g. we cannot poll a behaviour and if it returns something
-// of interest then call a method on another behaviour. We can do this in another wrapper
+// of interest then call a method on another behaviour. We can do this in the a wrapper
 // where we manually implement `NetworkBehaviour`, or the outer service where we drive the
 // Swarm; there we are free to call any of the behaviours.
+
+impl<P: StoreParams> Behaviour<P> {
+    pub fn new<S>(
+        nc: NetworkConfig,
+        dc: DiscoveryConfig,
+        mc: MembershipConfig,
+        store: S,
+    ) -> Result<Self, ConfigError>
+    where
+        S: BitswapStore<Params = P>,
+    {
+        Ok(Self {
+            ping: Default::default(),
+            identify: identify::Behaviour::new(identify::Config::new(
+                "ipfs/0.1.0".into(),
+                nc.local_public_key(),
+            )),
+            discovery: discovery::Behaviour::new(nc.clone(), dc)?,
+            membership: membership::Behaviour::new(nc, mc)?,
+            content: content::Behaviour::new(store),
+        })
+    }
+}
--- a/ipld/resolver/src/behaviour/mod.rs
+++ b/ipld/resolver/src/behaviour/mod.rs
@@ -58,9 +58,9 @@ pub struct Behaviour<P: StoreParams> {
 
 // Unfortunately by using `#[derive(NetworkBehaviour)]` we cannot easily inspects events
 // from the inner behaviours, e.g. we cannot poll a behaviour and if it returns something
-// of interest then call a method on another behaviour. We can do this in the a wrapper
+// of interest then call a method on another behaviour. We can do this in yet another wrapper
 // where we manually implement `NetworkBehaviour`, or the outer service where we drive the
-// Swarm; there we are free to call any of the behaviours.
+// Swarm; there we are free to call any of the behaviours as well as the Swarm.
 
 impl<P: StoreParams> Behaviour<P> {
     pub fn new<S>(
--- a/ipld/resolver/src/behaviour/mod.rs
+++ b/ipld/resolver/src/behaviour/mod.rs
@@ -10,9 +10,9 @@ use libp2p::{
 };
 use libp2p_bitswap::BitswapStore;
 
-mod content;
-mod discovery;
-mod membership;
+pub mod content;
+pub mod discovery;
+pub mod membership;
 
 pub use discovery::Config as DiscoveryConfig;
 pub use membership::Config as MembershipConfig;
@@ -82,5 +82,17 @@ impl<P: StoreParams> Behaviour<P> {
             membership: membership::Behaviour::new(nc, mc)?,
             content: content::Behaviour::new(store),
         })
+    }
+
+    pub fn discovery_mut(&mut self) -> &mut discovery::Behaviour {
+        &mut self.discovery
+    }
+
+    pub fn membership_mut(&mut self) -> &mut membership::Behaviour {
+        &mut self.membership
+    }
+
+    pub fn content_mut(&mut self) -> &mut content::Behaviour<P> {
+        &mut self.content
     }
 }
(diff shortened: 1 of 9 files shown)


ipld/resolver/src/service.rs
```diff
@@ -1,17 +1,431 @@
...
+            discovery::Event::Disconnected(_, _) => {}
+        }
+    }
+
+    fn handle_membership_event(&mut self, event: membership::Event) {
+        match event {
+            membership::Event::Skipped(peer_id) => {
+                // Don't repeatedly look up peers we can't add to the routing table.
+                if self.background_lookup_filter.insert(&peer_id) {
+                    self.discovery_mut().background_lookup(peer_id)
```
adlrocha, line 275: Are we refreshing the bloom filter after some time? I don´t know if having a lot of peers after some point could lead to skipping legit peers.

ipld/resolver/src/service.rs
```diff
@@ -1,17 +1,431 @@
...
+        error!("error sending resolve result; listener closed")
+    }
+}
+
+/// Builds the transport stack that libp2p will communicate over.
+///
+/// Based on the equivalent in Forest.
+pub fn build_transport(local_key: Keypair) -> Boxed<(PeerId, StreamMuxerBox)> {
+    let tcp_transport =
+        || libp2p::tcp::tokio::Transport::new(libp2p::tcp::Config::new().nodelay(true));
```
adlrocha, line 404: I don't know if it is already supported in rust-libp2p, but I would love if we could switch to quic in the future and only fallback to tcp if not supported :)

ipld/resolver/src/service.rs
```diff
@@ -1,17 +1,431 @@
...
+                    error!("error removing provided subnet: {e}")
+                }
+            }
+            Request::PinSubnet(id) => self.membership_mut().pin_subnet(id),
+            Request::UnpinSubnet(id) => self.membership_mut().unpin_subnet(&id),
+
+            Request::Resolve(cid, subnet_id, response_channel) => {
+                let peers = self.membership_mut().providers_of_subnet(&subnet_id);
+                if peers.is_empty() {
+                    send_resolve_result(response_channel, Err(anyhow!(NoKnownPeers(subnet_id))));
```
adlrocha, line 350: This would mean that no membership was picked up through gossipsub for the subnet, right? We could include in the future a fallback that optimistically asks our direct connections what subnets they are part of for the case we are lucky and we find a provider.

adlrocha (approved):
Overall, LGTM, but I am having a hard time putting all things together and understanding when each of the behaviors kick in. Maybe it is worth having a quick sync once the whole protocol is done for you to guide us through the code 🙏

Also, up till now we are using vanilla Bitswap, right? Which assumes that the content being retrieved lives in peers' BitswapStore, which is not currently the case for the agent.

ipld/resolver/src/behaviour/mod.rs
```diff
@@ -29,22 +34,65 @@ impl NetworkConfig {
...
+        mc: MembershipConfig,
+        store: S,
+    ) -> Result<Self, ConfigError>
+    where
+        S: BitswapStore<Params = P>,
+    {
+        Ok(Self {
+            ping: Default::default(),
+            identify: identify::Behaviour::new(identify::Config::new(
+                "ipfs/0.1.0".into(),
```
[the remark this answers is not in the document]
aakoshh: I don't know, I copied it as-is from Forest. The `KademliaConfig` says the protocol name can be used to distinguish different groups of peers, but the `IdentifyConfig` just says:
```
    /// Application-specific version of the protocol family used by the peer,
    /// e.g. `ipfs/1.0.0` or `polkadot/1.0.0`.
    pub protocol_version: String,
```
I fixed the `0.1.0` to be `1.0.0`, it looked like a typo. I suspect it doesn't matter what it is, it could be `ipc`? I can't see it used, but I only looked at the `Identify` behaviour.

For the record when my nodes connect to each other, I see these protocol being supported:
```
["/ipfs/ping/1.0.0", "/ipfs/id/1.0.0", "/ipfs/id/push/1.0.0", "/ipc/smoke-test/kad/1.0.0", "/meshsub/1.1.0", "/meshsub/1.0.0", "/ipfs-embed/bitswap/1.0.0"]
```

So this exact thing doesn't even appear :shrug:

ipld/resolver/src/service.rs
```diff
@@ -1,17 +1,431 @@
...
+                    None => { break; },
+                },
+                request = self.request_rx.recv() => match request {
+                    // A Client sent us a request.
+                    Some(req) => self.handle_request(req),
+                    // All Client instances have been dropped.
+                    // We could keep the Swarm alive to serve content to others,
+                    // but we ourselves are unable to send requests. Let's treat
+                    // this as time to quit.
+                    None => { break; }
```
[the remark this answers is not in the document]
aakoshh: In order to spawn a client, the `Service` would have to have a copy of the `Sender<Request>` channel (which is what happens in Forest. But the `Service` instance is consumed by `run`, so you can't call anything on it any more, you can't have a reference to it, you would not be able to clone a new `Sender`. And by itself having a copy of `Sender`, it would never experience this condition either, unless it manually drop it. 

So I decided that a cleaner way is for `Service::new` to return the `Service` and the `Sender` (wrapped into a `Client`). Then you can make as many clones of the `Client` as you want, before or after you `run` the `Service`. 

But you are right, you have to make sure not to drop all `Client` instances. But this is easy, because if you did, and you needed another one, you wouldn't be able to get it from anywhere, so the compiler wouldn't let this happen. And if you never ever need to make a `Client` call (after telling it what subnets you support), then yeah, you just need to make sure you keep an instance in your application state anyway.

ipld/resolver/src/service.rs
aakoshh: I am also thinking about this, but I haven't come up with a nice way to do it yet. A fresh start every now and then seems like a blunt and easy way, but leads to a burst of lookups after. Maybe a better protection would be to use a cache like the `guava` library supported in the JVM: it has limits on memory and a TTL as well.

ipld/resolver/src/service.rs
aakoshh: I will check. FWIW in #64 I factor out transport creation, so we can switch easily.

ipld/resolver/src/service.rs
aakoshh: Yes, most likely we asked too early and we should come back later, or perhaps find a known peer and add it to the static address list. 

I am also thinking on how to make the gossip better for newcomers, to lessen the trade-off of periodic republishing and let newcomers learn the providers ASAP.

Here's an idea: when we receive a gossip from a provider we haven't known about before, we publish our own in return. They must be a newcomer, so they could use our own information for sure. Perhaps with some cool-off period, like no repeated publishing within 10 seconds (just above the gossipsub broadcast time frame).

aakoshh:
> Also, up till now we are using vanilla Bitswap, right? Which assumes that the content being retrieved lives in peers' BitswapStore, which is not currently the case for the agent.

That's right, this Bitswap library interacts with the store. The agent will have to provide an implementation of the store that is fit for purpose:
* For requests, it would look in its own memory, or the Lotus node it is connected to 
* For respones, it would store the data in a `MemoryBlockstore`, so that it can serve it to Lotus later 

With Fendermint they can share the actual database, so this is easier. The potential question is whether databases of the same validator across the different levels of the hierarchy could be shared.

ipld/resolver/src/service.rs
aakoshh: Created https://github.com/consensus-shipyard/ipc-agent/issues/67

aakoshh:
I know you had a different idea of how to resolve content, by going through the Gateway actor itself, something like "give the the contents of the checkpoint identified by this CID", and the Gateway would collect every message that belongs to the checkpoint. Whereas Bitswap would go CID by CID until the whole thing is pulled. 

So, I reckon your implementation of the `BitswapStore` would have to be special in that it would first ask the Gateway actor for the CIDs of the checkpoints and load them into its memory; when a request comes, it's only served if it was discovered this way, otherwise it's considered not asking for a part of a checkpoint. 

I recognise that this may not work well for you if you want to respond to a CID by some data that doesn't actually correspond to that CID, which would be impossible, so I urge you to find a compromise.

ipld/resolver/src/behaviour/mod.rs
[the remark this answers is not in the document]
adlrocha: I see they are using it for authentication purposes.

adlrocha:
> I recognise that this may not work well for you if you want to respond to a CID by some data that doesn't actually correspond to that CID, which would be impossible, so I urge you to find a compromise.

This will never be the case. I looked at the BitswapStore implementation in the smoke test and for our case it may just suffice to add a wrapper for the async calls to the gateway or relevant subnet state.

PR #56, opened 2023-03-02
Create subnet cli

cryptoAtwill:
Create subnet actor command line handler.

This is just a direct json rpc call to ipc agent json rpc server.

Update: Add global config params so that config path will be shared by all cli commands.

### DIFF
--- a/src/cli/commands/create.rs
+++ b/src/cli/commands/create.rs
@@ -2,12 +2,14 @@
 // SPDX-License-Identifier: MIT
 //! Create subnet cli command handler.
 
+use anyhow::{anyhow, Result};
 use async_trait::async_trait;
 use clap::Args;
 use fvm_shared::clock::ChainEpoch;
 use std::fmt::Debug;
+use url::Url;
 
-use crate::cli::CommandLineHandler;
+use crate::cli::{CommandLineHandler, GlobalParams};
 use crate::config::json_rpc_methods;
 use crate::jsonrpc::{JsonRpcClient, JsonRpcClientImpl};
 use crate::server::{CreateSubnetParams, CreateSubnetResponse};
@@ -19,8 +21,11 @@ pub(crate) struct CreateSubnet;
 impl CommandLineHandler for CreateSubnet {
     type Arguments = CreateSubnetArgs;
 
-    async fn handle(arguments: &Self::Arguments) -> anyhow::Result<()> {
+    async fn handle(global: &GlobalParams, arguments: &Self::Arguments) -> anyhow::Result<()> {
         log::debug!("launching json rpc server with args: {:?}", arguments);
+
+        let url = get_ipc_agent_url(&arguments.ipc_agent_url, global, &arguments.parent)?;
+        let json_rpc_client = JsonRpcClientImpl::new(url, None);
 
         let params = CreateSubnetParams {
             parent: arguments.parent.clone(),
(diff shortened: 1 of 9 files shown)


src/cli/commands/create.rs
```diff
@@ -0,0 +1,68 @@
...
+        let params = CreateSubnetParams {
+            parent: arguments.parent.clone(),
+            name: arguments.name.clone(),
+            min_validator_stake: arguments.min_validator_stake,
+            min_validators: arguments.min_validators,
+            finality_threshold: arguments.finality_threshold,
+            check_period: arguments.check_period,
+        };
+
+        let url = arguments.ipc_agent_url.parse()?;
```
adlrocha, line 34: Can we fallback to the default one if this isn't set?

src/cli/commands/create.rs
```diff
@@ -0,0 +1,68 @@
...
+        let json_rpc_client = JsonRpcClientImpl::new(url, None);
```
adlrocha, line 35: We may need this for all other functions and we may be able to take it out to a common func.

src/config/server.rs
```diff
@@ -9,3 +9,7 @@ pub const JSON_RPC_ENDPOINT: &str = "json_rpc";
 pub struct Server {
     pub json_rpc_address: SocketAddr,
 }
+
+pub mod json_rpc_methods {
+    pub const CREATE_SUBNET: &str = "create_subnet";
```
adlrocha, line 14: Opened a poll in Slack to agree on what convention we want to follow. Let's wait for the results to change/keep this.

adlrocha (approved):
LGTM. Wait for the RPC naming convention to change it before merging. Thanks 🙏

src/cli/commands/create.rs
cryptoAtwill: do we have a default value? There is a config file, but the caller can specify any `ipc_agent_url`, should we just let the user config?

src/cli/commands/create.rs
cryptoAtwill: If I understand correctly, in this CLI, it's only a json rpc client initialization. The `JsonRpcClientImpl::new` is already the most simple form. Or are you referring to the initialization of LotusJsonRpcClient?

cryptoAtwill:
@adlrocha The naming conventions for json rpc methods are updated. Please take a look at other comments.

src/cli/commands/create.rs
adlrocha: We are setting a default path for config and optionally we can pass a config path also as argument.

adlrocha (approved):

src/cli/commands/create.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+fn get_ipc_agent_url(
+    ipc_agent_url: &Option<String>,
+    global: &GlobalParams,
+    subnet_str: &str,
+) -> Result<Url> {
+    let url = match ipc_agent_url {
+        Some(url) => url.parse()?,
+        None => {
```
adlrocha, line 79: I don't get what you are doing here, why are we getting the subnet url here? We should be getting the one from the agent not one for the subnet.

src/cli/mod.rs
```diff
@@ -22,5 +26,30 @@ pub trait CommandLineHandler {
...
+    )]
+    config_path: Option<String>,
+}
+
+impl GlobalParams {
+    pub fn config_path(&self) -> String {
+        self.config_path.clone().unwrap_or_else(|| {
+            let home = match std::env::var("HOME") {
+                Ok(home) => home,
+                Err(_) => String::from(""), // cannot get home, just use current dir
```
adlrocha, line 45: I don't think this would lead to an accurate value, right?

adlrocha (changes requested):
The new change for the config path LGTM, but have a look to my comments as I don't know if we introduced slight bugs.

src/cli/mod.rs
cryptoAtwill: as discussed, should panic!

adlrocha (approved):
Thank you for the changes, feel free to merge.

PR #61, opened 2023-03-03
Implements logic for actively checkpointing the state of managed subnets.

hmoniz:
This PR implements the checkpointing logic where the subnets specified in the config are actively monitored for checkpointing epochs and, for each account that the agent manages in a subnet, a checkpoint is submitted to the subnet actor in the parent subnet. PR #60 is a dependency.

### DIFF
--- a/src/manager/checkpoint.rs
+++ b/src/manager/checkpoint.rs
@@ -1,0 +1,231 @@
+// Copyright 2022-2023 Protocol Labs
+// SPDX-License-Identifier: MIT
+use std::collections::{HashMap, HashSet};
+use std::collections::hash_map::RandomState;
+use std::sync::Arc;
+use std::time::Duration;
+
+use anyhow::Result;
+use cid::Cid;
+use fil_actors_runtime::cbor;
+use futures_util::stream::FuturesUnordered;
+use futures_util::StreamExt;
+use fvm_shared::address::Address;
+use fvm_shared::clock::ChainEpoch;
+use fvm_shared::MethodNum;
+use ipc_gateway::Checkpoint;
+use ipc_sdk::subnet_id::SubnetID;
+use primitives::TCid;
+use tokio::select;
+use tokio::sync::Notify;
+use tokio::time::sleep;
+use tokio_graceful_shutdown::SubsystemHandle;
+
+use crate::config::{ReloadableConfig, Subnet};
+use crate::jsonrpc::JsonRpcClient;
+use crate::lotus::client::LotusJsonRPCClient;
+use crate::lotus::message::mpool::MpoolPushMessage;
+use crate::lotus::LotusClient;
+
+/// The frequency at which to check a new chain head.
+const CHAIN_HEAD_REQUEST_PERIOD: Duration = Duration::from_secs(10);
+
+/// The `CheckpointSubsystem`. When run, it actively monitors subnets and submits checkpoints.
+struct CheckpointSubsystem {
+    /// The subsystem uses a `ReloadableConfig` to ensure that, at all, times, the subnets under
+    /// management are those in the latest version of the config.
+    config: ReloadableConfig,
+}
+
+impl CheckpointSubsystem {
+    /// Creates a new `CheckpointSubsystem` with a configuration `config`.
+    #[allow(dead_code)]
+    fn new(config: ReloadableConfig) -> Self {
+        Self { config }
+    }
+
+    /// Runs the checkpoint subsystem, which actively monitors subnets and submits checkpoints.
+    /// For each (account, subnet) that exists in the config, the subnet is monitored and checkpoints
+    /// are submitted at the appropriate epochs.
+    #[allow(dead_code)]
+    async fn run(&self, subsys: SubsystemHandle) -> Result<()> {
+        // Each event in this channel is notification of a new config.
+        let mut config_chan = self.config.new_subscriber();
+
+        loop {
+            // Load the latest config.
+            let config = self.config.get_config();
+
+            // Create a `manage_subnet` future for each (child, parent) subnet pair under management
+            // and collect them in a `FuturesUnordered` set.
+            let manage_subnet_futures = FuturesUnordered::new();
+            let stop_subnet_managers = Arc::new(Notify::new());
+            for (child, parent) in subnets_to_manage(&config.subnets) {
+                manage_subnet_futures
+                    .push(manage_subnet((child, parent), stop_subnet_managers.clone()));
+            }
+
+            // Spawn a task to drive the `manage_subnet` futures.
+            let manage_subnets_task =
+                tokio::spawn(manage_subnet_futures.collect::<Vec<Result<()>>>());
+
+            // Watch for shutdown requests and config changes.
+            let is_shutdown = select! {
+                _ = subsys.on_shutdown_requested() => { true },
+                r = config_chan.recv() => {
+                    match r {
+                        Ok(_) => { false },
+                        Err(_) => { true },
+                    }
+                },
+            };
+
+            if is_shutdown {
+                // Cleanly stop the `manage_subnet` futures and return.
+                stop_subnet_managers.notify_waiters();
+                let results = manage_subnets_task.await?;
+                results.into_iter().collect::<Result<Vec<_>>>()?;
+                return anyhow::Ok(());
+            }
+        }
+    }
+}
+
+/// This function takes a `HashMap<String, Subnet>` and returns a `Vec` of tuples of the form
+/// `(child_subnet, parent_subnet)`, where `child_subnet` is a subnet that we need to actively
+/// manage checkpoint for. This means that for each `child_subnet` there exists at least one account
+/// for which we need to submit checkpoints on behalf of to `parent_subnet`, which must also be
+/// present in the map.
+fn subnets_to_manage(subnets: &HashMap<String, Subnet>) -> Vec<(Subnet, Subnet)> {
+    // First, we remap subnets by SubnetID.
+    let subnets_by_id: HashMap<SubnetID, Subnet> = subnets
+        .values()
+        .map(|s| (s.id.clone(), s.clone()))
+        .collect();
+
+    // Then, we filter for subnets that have at least one account and for which the parent subnet
+    // is also in the map, and map into a Vec of (child_subnet, parent_subnet) tuples.
+    subnets_by_id
+        .values()
+        .filter(|s| !s.accounts.is_empty())
+        .filter(|s| s.id.parent().is_some() && subnets_by_id.contains_key(&s.id.parent().unwrap()))
+        .map(|s| (s.clone(), subnets_by_id[&s.id.parent().unwrap()].clone()))
+        .collect()
+}
+
+/// Monitors a subnet `child` for checkpoint blocks. It emits an event for every new checkpoint block.
+async fn manage_subnet((child, parent): (Subnet, Subnet), stop_notify: Arc<Notify>) -> Result<()> {
+    let child_client = LotusJsonRPCClient::from_subnet(&child);
+    let parent_client = LotusJsonRPCClient::from_subnet(&parent);
+
+    // Read the parent's chain head and obtain the tip set CID.
+    let parent_head = parent_client.chain_head().await?;
+    let cid_map = parent_head.cids.first().unwrap().clone();
+    let parent_tip_set = Cid::try_from(cid_map)?;
+
+    // Extract the checkpoint period from the state of the subnet actor in the parent.
+    let state = parent_client
+        .ipc_read_subnet_actor_state(parent_tip_set)
+        .await?;
+    let period = state.check_period;
+
+    // We should have a way of knowing whether the validator has voted in the current open
+    // checkpoint epoch.
+    // TODO: Hook this up to the new IPC methods.
+
+    // We can now start looping. In each loop we read the child subnet's chain head and check if
+    // it's a checkpoint epoch. If it is, we construct and submit a checkpoint.
+    loop {
+        let child_head = child_client.chain_head().await?;
+        let epoch: ChainEpoch = ChainEpoch::try_from(child_head.height)?;
+        if epoch % period == 0 {
+            // It's a checkpointing epoch and we may have checkpoints to submit.
+
+            // First, we check which accounts are in the validator set. This is done by reading
+            // the parent's chain head and requesting the state at that tip set.
+            let parent_head = parent_client.chain_head().await?;
+            // A key assumption we make now is that each block has exactly one tip set. We panic
+            // if this is not the case as it violates our assumption.
+            // TODO: update this logic once the assumption changes (i.e., mainnet)
+            assert_eq!(parent_head.cids.len(), 1);
+            let cid_map = parent_head.cids.first().unwrap().clone();
+            let parent_tip_set = Cid::try_from(cid_map)?;
+
+            let subnet_actor_state = parent_client
+                .ipc_read_subnet_actor_state(parent_tip_set)
+                .await?;
+            let validator_set: HashSet<Address, RandomState> = HashSet::from_iter(
+                subnet_actor_state
+                    .validator_set
+                    .validators()
+                    .iter()
+                    .map(|v| v.addr),
+            );
+
+            // Now, for each account defined in the `child` subnet that is in the validator set, we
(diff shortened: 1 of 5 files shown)


hmoniz:
Current dependencies on/for this PR:
* main
  * **PR #60** <a href="https://app.graphite.dev/github/pr/consensus-shipyard/ipc-agent/60" target="_blank">Graphite</a> 
    * **PR #61** <a href="https://app.graphite.dev/github/pr/consensus-shipyard/ipc-agent/61" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/consensus-shipyard/ipc-agent/61?utm_source=stack-comment).

src/manager/checkpoint.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/// This function takes a `HashMap<String, Subnet>` and returns a `Vec` of tuples of the form
+/// `(child_subnet, parent_subnet)`, where `child_subnet` is a subnet that we need to actively
+/// manage checkpoint for. This means that for each `child_subnet` there exists at least one account
+/// for which we need to submit checkpoints on behalf of to `parent_subnet`, which must also be
+/// present in the map.
+fn subnets_to_manage(subnets: HashMap<String, Subnet>) -> Vec<(Subnet, Subnet)> {
+    // First, we remap subnets by SubnetID.
+    let subnets_by_id: HashMap<SubnetID, Subnet> = subnets
```
cryptoAtwill, line 45: Just wondering, in other json rpc methods, converting to `HashMap<SubnetID, Subnet>` is also needed. Not sure if we can just update the config to `HashMap<SubnetID, Subnet>` instead of `Hashmap<String, Subnet>`

src/manager/checkpoint.rs
```diff
@@ -0,0 +1,160 @@
...
+/// Starts the checkpoint manager.
+#[allow(dead_code)]
+pub async fn start(subnets: HashMap<String, Subnet>) {
+    let managed_subnets = subnets_to_manage(subnets);
+
+    let mut futures = Vec::new();
+    for (child, parent) in managed_subnets {
+        futures.push(manage_subnet(child, parent));
+    }
+    join_all(futures).await;
```
cryptoAtwill, line 35: Not sure how relevant this comment is, but if we use `join_all` then this `start` is cannot be cancelled nor changed and it will run forever. Not sure if this will ever be subject to change.

cryptoAtwill (changes requested):

src/manager/checkpoint.rs
adlrocha: This is a good point, but IIRC there was a reason why this couldn't be done. Let's wait for @hmoniz's input.

src/manager/checkpoint.rs
```diff
@@ -0,0 +1,160 @@
...
+/// for which we need to submit checkpoints on behalf of to `parent_subnet`, which must also be
+/// present in the map.
+fn subnets_to_manage(subnets: HashMap<String, Subnet>) -> Vec<(Subnet, Subnet)> {
+    // First, we remap subnets by SubnetID.
+    let subnets_by_id: HashMap<SubnetID, Subnet> = subnets
+        .values()
+        .map(|s| (s.id.clone(), s.clone()))
+        .collect();
+
+    // Then, we filter for subnets that have at least one account and for which the parent subnet
```
adlrocha, line 50: Should we also add a `validator` field in the config for child subnets to flag the subnets that an IPC agent is part of as validator? I think this filter is useful, but it may also be the case that an IPC agent is syncing with a parent-child pair where it is not participating as a validator and it doesn't need to run the checkpoint process.

src/manager/checkpoint.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/// Monitors a subnet `child` for checkpoint blocks. It emits an event for every new checkpoint block.
+async fn manage_subnet(child: Subnet, parent: Subnet) -> Result<()> {
+    let child_client = LotusJsonRPCClient::from_subnet(&child);
+    let parent_client = LotusJsonRPCClient::from_subnet(&parent);
+
+    // Read the parent's chain head and obtain the tip set CID.
+    let parent_head = parent_client.chain_head().await?;
+    let cid_map = parent_head.cids.last().unwrap().clone();
```
adlrocha, line 67: nit: can we get `first()` here? we shouldn't ever see more than one block, but in certain parts of the lotus code we use `block[0]`, that way if something goes wrong we are sure that we are being consistent with the lotus logic.

src/manager/checkpoint.rs
```diff
@@ -0,0 +1,160 @@
...
+    let parent_tip_set = Cid::try_from(cid_map)?;
+
+    // Extract the checkpoint period from the state of the subnet actor in the parent.
+    let state = parent_client
+        .ipc_read_subnet_actor_state(parent_tip_set)
+        .await?;
+    let period = state.check_period;
+
+    // We should have a way of knowing whether the validator has voted in the current open
```
adlrocha, line 76: We can read this information from the subnet actor state. Let me know if you want me to provide you with a convenient access method.

src/manager/checkpoint.rs
```diff
@@ -0,0 +1,160 @@
...
+        let epoch: ChainEpoch = ChainEpoch::try_from(child_head.height)?;
+        if epoch % period == 0 {
+            // It's a checkpointing epoch and we may have checkpoints to submit.
+
+            // First, we check which accounts are in the validator set. This is done by reading
+            // the parent's chain head and requesting the state at that tip set.
+            let parent_head = parent_client.chain_head().await?;
+            // A key assumption we make now is that each block has exactly one tip set. We panic
+            // if this is not the case as it violates our assumption.
+            assert_eq!(parent_head.cids.len(), 1);
```
adlrocha, line 93: This is the case for Mir, once we move to Filecoin mainnet as the root for M3 this won't be the case. Let's maybe add a TODO here to remind ourselves. Thanks.

src/manager/checkpoint.rs
```diff
@@ -0,0 +1,160 @@
...
+    // it's a checkpoint epoch. If it is, we construct and submit a checkpoint.
+    loop {
+        let child_head = child_client.chain_head().await?;
+        let epoch: ChainEpoch = ChainEpoch::try_from(child_head.height)?;
+        if epoch % period == 0 {
+            // It's a checkpointing epoch and we may have checkpoints to submit.
+
+            // First, we check which accounts are in the validator set. This is done by reading
+            // the parent's chain head and requesting the state at that tip set.
+            let parent_head = parent_client.chain_head().await?;
```
adlrocha, line 90: Double-check: We assume here that the endpoint to the client won't change through the lifecycle of the ipc agent, right? I think this is a fair assumption for now, a user can restart if it needs to load a new config.

adlrocha (commented):
Part 1 of my review. Graphite UI is killing my comments with every update to the PR so persisting these to avoid losing them.

src/manager/checkpoint.rs
```diff
@@ -0,0 +1,160 @@
...
+    let period = state.check_period;
+
+    // We should have a way of knowing whether the validator has voted in the current open
+    // checkpoint epoch. For now, we maintain this information in memory and upon a restart only
+    // vote for the next epoch.
+
+    // We can now start looping. In each loop we read the child subnet's chain head and check if
+    // it's a checkpoint epoch. If it is, we construct and submit a checkpoint.
+    loop {
+        let child_head = child_client.chain_head().await?;
```
adlrocha, line 83: I don't know if I like this check. What if you miss the epoch of the checkpoint when reading the head? I would suggest the following change. 
- Read the height of the head. 
- If head > epoch for the next checkpoint, see if we already voted for that checkpoint. 
- If this is not the case, submit the checkpoint. 
I think this may cover us against certain rough edges,

src/manager/checkpoint.rs
adlrocha: Wait, inside the loop you don't create the client `from_subnet` but you do inside `submit_checkpoint` what may lead to inconsistencies. Am I missing something?

adlrocha (changes requested):
Some clarifications needed, but we are close. Thanks!

src/manager/checkpoint.rs
hmoniz: In the config, each subnet is specified through a section declaration of the form `[subnet.C]`, where `C` is a human-readable name that we attribute to the subnet. The `C` is what ends up being the key for the `HashMap`. I, too, would prefer for the map to be of the form `<SubnetID, Subnet>` right off the bat. But for this to happen, we would have to instead declare each subnet section as, for instance, `[subnet./root/f0100]` and write some custom deserialization logic for it. I'm not sure if this is worth the loss in UX. Maybe it's better to simply have a convenience function that remaps from `<String, Subnet>` to `<SubnetID, Subnet`>. Wdyt?

src/manager/checkpoint.rs
hmoniz: I don't know. I think that will add unnecessary burden to the user and potentially make them miss checkpoints as they would have to manually ensure that the config is in sync with the subnet actor state.

src/manager/checkpoint.rs
hmoniz: I can do that, of course. But isn't the last block the one that represents the state we want to check for?

src/manager/checkpoint.rs
hmoniz: That would be nice as it would make this logic more robust!

src/manager/checkpoint.rs
hmoniz: Yes. Particularly, if we have the method you suggest above, it would make this relatively simple to implement.

src/manager/checkpoint.rs
hmoniz: I'll add a TODO. The reason I set this panic is precisely because it should blow up to reminds that the logic here needs to be updated.

src/manager/checkpoint.rs
hmoniz: For now, this implementation assumes a static config. Once we have some form of hot reloading, this code will need to be restructured. In any case, for `submit_checkpoint` we should probably be passing a reference to client already constructed instead of instantiating a new one. I'll change that.

hmoniz: Actually, I'll wait for the hot config PR before making any changes here as that will imply a deeper restructuring of this code.

src/manager/checkpoint.rs
hmoniz: Once we have the config reloading in place, we will want to "unwatch" subnets and start watching new nodes. This whole code will need to be restructured to accommodate that.

src/manager/checkpoint.rs
cryptoAtwill: this works as well!

src/manager/checkpoint.rs
adlrocha: You are right, I saw afterwards that you are explicitly checking if we hold a validator address in the subnet, so no need for this. I forgot to remove this comment, resolving.

src/manager/checkpoint.rs
adlrocha: But this list of CIDs is the list of blocks for the heaviest height, right? So for one block per tipset `first() == last()` or am I wrong?

src/manager/checkpoint.rs
adlrocha: I will add an endpoint for that in Lotus, I'll ping you when is done.

src/manager/checkpoint.rs
adlrocha: SGTM. I like the approach.

src/manager/checkpoint.rs
adlrocha: When you tackle the code reestructure (wither in this PR or a follow-up) maybe is worth leveraging the pool abstraction that @cryptoAtwill has introduced here: https://github.com/consensus-shipyard/ipc-agent/pull/71/files#diff-17a90fa412b940e0b71c1f95f45bc9e3bbd5a05cf6793320273774d15a3879e7R32

src/manager/checkpoint.rs
hmoniz: My previous comment was a way of asking whether we should change this to `first()` or change lotus to use `last()` (or equivalent). If the last block in the tipset is the one that represents the state we're considering, then it should be `last()`. In any case, change this here for now.

src/manager/checkpoint.rs
adlrocha: Let's go with this approach of remapping the hashmap, and let's see if users get confused or not with the current config format and change accordingly.

src/manager/checkpoint.rs
hmoniz: I updated this logic. The driving of the futures was pushed to a dedicated task and I've added a select to drive the check of shutdown signal and config changes.

src/manager/checkpoint.rs
```diff
@@ -0,0 +1,196 @@
...
+
+        let manage_subnet_futures = FuturesUnordered::new();
+        for (child, parent) in managed_subnets {
+            //manage_subnet_futures.push(async { 1_i32 });
+            manage_subnet_futures.push(manage_subnet((child, parent), stop_notify.clone()));
+        }
+
+        // If we receive a new config, we stop all `manage_subnet` futures and continue to the next
+        // iteration where the new config will be loaded.
+        let (_, results): (Result<()>, _) = join!(
```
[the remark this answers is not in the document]
hmoniz: This was revamped. Please take another look.

src/manager/checkpoint.rs
hmoniz: I'll address this in a follow up PR to not pollute this one much more.

src/manager/checkpoint.rs
hmoniz: Added a TODO to address this right after this PR.

adlrocha (approved):
Apart from the few TODOs that you added for follow-up PRs this LGTM. We still have to make the CI happy.

Additionally, we should integrate it in the daemon so it is started with the IPC agent, right?

src/manager/checkpoint.rs
```diff
@@ -0,0 +1,228 @@
...
+            // TODO: update this logic once the assumption changes (i.e., mainnet)
+            assert_eq!(parent_head.cids.len(), 1);
+            let cid_map = parent_head.cids.first().unwrap().clone();
+            let parent_tip_set = Cid::try_from(cid_map)?;
+
+            let subnet_actor_state = parent_client
+                .ipc_read_subnet_actor_state(parent_tip_set)
+                .await?;
+            let validator_set: HashSet<Address, RandomState> =
+                HashSet::from_iter(subnet_actor_state.validator_set.iter().map(|v| v.addr));
```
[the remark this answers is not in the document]
hmoniz: Done

src/manager/checkpoint.rs
```diff
@@ -0,0 +1,228 @@
...
+    // Get the CID of previous checkpoint of the child subnet from the gateway actor of the parent
+    // subnet.
+    let response = parent_client
+        .ipc_get_prev_checkpoint_for_child(child_subnet.id.clone())
+        .await?;
+    let cid = Cid::try_from(response.cid)?;
+    checkpoint.data.prev_check = TCid::from(cid);
+
+    // The "tip_set".
+    checkpoint.data.tip_set = child_tip_set.to_bytes();
```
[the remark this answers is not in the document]
hmoniz: Updated

src/manager/checkpoint.rs
```diff
@@ -0,0 +1,228 @@
...
+use crate::lotus::message::mpool::MpoolPushMessage;
+use crate::lotus::LotusClient;
+
+/// The frequency at which to check a new chain head.
+const CHAIN_HEAD_REQUEST_PERIOD: Duration = Duration::from_secs(10);
+
+/// The `CheckpointSubsystem`. When run, it actively monitors subnets and submits checkpoints.
+struct CheckpointSubsystem {
+    /// The subsystem uses a `ReloadableConfig` to ensure that, at all, times, the subnets under
+    /// management are those in the latest version of t he config.
```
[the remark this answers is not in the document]
hmoniz: Updated

cryptoAtwill (approved):

PR #62, opened 2023-03-03
IPC-58: Disable Kademlia storage inserts

### LINKED ISSUE #38: IPLD Resolver: Integration Test
Part of #17 

Start an in-memory cluster of #37 and run tests on them to show that they discover each other, learn the membership table, and are able to resolve CIDs. Run scenarios for late joiners to see how Gossipsub behaves.

### LINKED ISSUE #58: IPLD Resolver: Disable Kademlia content storage
As described in https://adlrocha.substack.com/p/adlrocha-beyond-bitswap-i , IPFS uses Kademlia to store a list of peers who can serve a piece of content, and not just a list of peers. It's only by accident that some blockchains decided to start using Kademlia for peer discovery, and now (at least for me) that's how I mostly think about it, but the `kad` protocol in `libp2p` is full implementation and thus supports more kinds of interaction than just peer lookup. 

In particular, there are [AddProvider and PutRecord](https://github.com/libp2p/rust-libp2p/blob/v0.50.0/protocols/kad/src/handler.rs#L306-L347) which allow peers to register themselves as hosts and also to put data into our node. This is a potential attack vector, as malicious users could put pressure on our memory. 

These events are handled in [on_connection_handler_event](https://github.com/libp2p/rust-libp2p/blob/v0.50.0/protocols/kad/src/behaviour.rs#L2018) and if we look at the [PutRecord handler](https://github.com/libp2p/rust-libp2p/blob/v0.50.0/protocols/kad/src/behaviour.rs#L2250) we can see that whether anything gets stored depends on the [record_filtering](https://github.com/libp2p/rust-libp2p/blob/v0.50.0/protocols/kad/src/behaviour.rs#L1673) setting. By using `KademliaStoreInserts::FilterBoth` we only get an event, while `KademliaStoreInserts::Unfiltered` puts it in the store first. `Unfiltered` happens to be the [default setting](https://github.com/libp2p/rust-libp2p/blob/v0.50.0/protocols/kad/src/behaviour.rs#L191). 

To prevent anyone from storing records, we can either use the `FilterBoth` setting, or we can stop these events from reaching the `Kademlia` behaviour in our `discovery::Behaviour::on_connection_event_handler`.

aakoshh:
Closes #58 

The PR disables the storing of Kademlia records in the store and warns if any peers sends us such requests.

### DIFF
--- a/ipld/resolver/src/behaviour/discovery.rs
+++ b/ipld/resolver/src/behaviour/discovery.rs
@@ -12,8 +12,8 @@ use std::{
 use libp2p::{
     core::connection::ConnectionId,
     kad::{
-        handler::KademliaHandlerProto, store::MemoryStore, Kademlia, KademliaConfig, KademliaEvent,
-        QueryId,
+        handler::KademliaHandlerProto, store::MemoryStore, InboundRequest, Kademlia,
+        KademliaConfig, KademliaEvent, KademliaStoreInserts, QueryId,
     },
     multiaddr::Protocol,
     swarm::{
@@ -24,7 +24,7 @@ use libp2p::{
     },
     Multiaddr, PeerId,
 };
-use log::debug;
+use log::{debug, warn};
 use tokio::time::Interval;
 
 use super::NetworkConfig;
@@ -114,6 +114,10 @@ impl Behaviour {
             let mut kad_config = KademliaConfig::default();
             let protocol_name = format!("/ipc/{}/kad/1.0.0", nc.network_name);
             kad_config.set_protocol_names(vec![Cow::Owned(protocol_name.as_bytes().to_vec())]);
+
+            // Disable inserting records into the memory store, so peers cannot send `PutRecord`
+            // messages to store content in the memory of our node.
+            kad_config.set_record_filtering(KademliaStoreInserts::FilterBoth);
 
             let store = MemoryStore::new(nc.local_peer_id());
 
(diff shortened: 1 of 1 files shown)


ipld/resolver/src/behaviour/discovery.rs
```diff
@@ -115,6 +115,10 @@ impl Behaviour {
             let protocol_name = format!("/ipc/{}/kad/1.0.0", nc.network_name);
             kad_config.set_protocol_names(vec![Cow::Owned(protocol_name.as_bytes().to_vec())]);
 
+            // Disable inserting records into the memory store, so peers cannot send `PutRecord`
+            // messages to store content in the memory of our node.
+            kad_config.set_record_filtering(KademliaStoreInserts::FilterBoth);
```
adlrocha, line 120: Just a sanity-check (I haven't looked in-depth to the rust impl): disabling `StoreInserts` still allow peers to advertise their contact information on a `PeerRecord` right? If not this may break our discovery service.

adlrocha (approved):
Have a look at my comment (just to double-check we don't break something(. If this is not the case LGTM.

ipld/resolver/src/behaviour/discovery.rs
aakoshh: It seems to work. `find_nearest_peers`, `bootstrap`, `AddProvider` and `PutRecord` all enqueue background queries to find the `ClosestPeers` to the one in the record, so they have the same effect. It looks to me like if we don't want to advertise nodes providing specific content, then we can disable the store completely.

PR #63, opened 2023-03-03
Reload config

cryptoAtwill:
Update config while the json rpc server is running.

`ReloadableConfig` exposes the latest config through `get_config` method. Use this you will always the latest config. At the same time, it also exposes `new_subscriber`. If caller needs to be notified when config has updated, just make a new subscription. Once received a notification, then just read the config again to obtain the latest config.

See `reload_works` test case for sample usage.

### DIFF
--- a/src/config/reload.rs
+++ b/src/config/reload.rs
@@ -1,3 +1,5 @@
+// Copyright 2022-2023 Protocol Labs
+// SPDX-License-Identifier: MIT
 //! Reloadable config
 
 use crate::config::Config;
--- a/src/config/reload.rs
+++ b/src/config/reload.rs
@@ -3,27 +3,15 @@
 //! Reloadable config
 
 use crate::config::Config;
-use anyhow::{anyhow, Result};
+use anyhow::Result;
 use std::ops::DerefMut;
 use std::path::Path;
 use std::sync::{Arc, RwLock};
-use tokio::sync::{broadcast, mpsc, oneshot};
-
-/// Internal request for reloadable config
-#[derive(Debug)]
-enum Request {
-    Reload {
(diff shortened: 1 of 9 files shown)


src/config/reload.rs
```diff
@@ -0,0 +1,57 @@
...
+use crate::config::Config;
+use anyhow::Result;
+use std::ops::DerefMut;
+use std::path::Path;
+use std::sync::{Arc, RwLock};
+use tokio::sync::broadcast;
+
+/// Reloadable configuration.
+pub struct ReloadableConfig {
+    config: Arc<RwLock<Arc<Config>>>,
```
adlrocha, line 14: fix?
```suggestion
    config: RwLock<Arc<Config>>,
```

src/config/reload.rs
```diff
@@ -0,0 +1,57 @@
...
+    broadcast_tx: broadcast::Sender<()>,
+    #[allow(dead_code)]
+    broadcast_rx: broadcast::Receiver<()>,
```
adlrocha, line 17: Add a note that we keep it to avoid dropping it but not used for the moment.

src/config/tests.rs
```diff
@@ -0,0 +1,184 @@
...
+        .to_os_string()
+        .into_string()
+        .unwrap();
+
+    file.write_all(config_str.as_bytes()).unwrap();
+
+    let h = Arc::new(ReloadableConfig::new(path.clone()).unwrap());
+    let original_config = h.get_config();
+
+    let pair = Arc::new((Mutex::new(false), Condvar::new()));
```
adlrocha, line 44: I think it'll help if you add a quick comment describing what you are doing here with the lock and the atomic var (in the end gow you imlpemented the barrier) .

PR #64, opened 2023-03-03
IPC-38: IPLD Resolver smoke test

### LINKED ISSUE #38: IPLD Resolver: Integration Test

aakoshh:
Closes #38 

Smoke test exercising the simplest possible scenario of starting some nodes, bootstrapping from the 1st, putting some content in the store of the 2nd, resolving by the 3rd.

Example run: 
```console
RUST_LOG=debug cargo test -p ipc_ipld_resolver --test smoke
```

There are some non-ideal things left in the PR, to be fixed later:
* it would be nice to switch to structured logging
* it would be great not to have to rely on sleep in the tests

But I think this is not bad to show that something works, with all the fixes that needed for it to happen, it's not a small PR.


ipld/resolver/src/behaviour/discovery.rs
```diff
@@ -165,6 +173,26 @@ impl Behaviour {
     fn is_static(&self, peer_id: PeerId) -> bool {
         self.static_addresses.iter().any(|(id, _)| *id == peer_id)
     }
+
+    /// Add addresses we learned from the `Identify` protocol to Kademlia.
+    ///
+    /// This seems to be the only way, because Kademlia rightfully treats
+    /// incoming connections as ephemeral addresses, but doesn't have an
+    /// alternative exchange mechanism.
+    pub fn add_identified(&mut self, peer_id: &PeerId, info: &Info) {
```
aakoshh, line 182: TODO: https://github.com/consensus-shipyard/ipc-agent/issues/70

aakoshh: Done in #74

ipld/resolver/tests/smoke.rs
```diff
@@ -0,0 +1,250 @@
...
+    let subnet_id = make_subnet_id(1001);
+
+    cluster.agents[provider_idx]
+        .client
+        .add_provided_subnet(subnet_id.clone())
+        .expect("failed to add provided subnet");
+
+    // Wait a little for the gossip to spread and peer lookups to happen, then another round of gossip.
+    // TODO: Wait on some condition instead of sleep.
+    tokio::time::sleep(Duration::from_secs(2)).await;
```
adlrocha, line 147: This have been shown to be a recurring issue when testing gossipsub-related protocols. >1s sync delay (which is the default heartbeat time) is the easy one, but IIRC we can probably listen or poll the libp2p host to see if it is bootrstapped. I don't know if this can be done easily in Rust, I did this once in Go.

(We can do something similar with the bootstrap delay above)

adlrocha (approved):
Amazing work! It is so exciting that this works :raised_hands:

ipld/resolver/tests/smoke.rs
aakoshh: Thank you for the tips. I can also add more events to show when things have progressed, and I have seen examples in the libp2p tests of a combination of a polling function and a timeout.

PR #71, opened 2023-03-06
Create subnet with reloading config

cryptoAtwill:
This PR builds upon config reloading, #63. 

When creating the json rpc subnet, if the target parent is not found, it will checks against the latest config. If the new subnet config is present, it would dynamically create the new subnet manager and add to the pool of connections.


src/config/mod.rs
```diff
@@ -25,7 +25,7 @@ pub use server::JSON_RPC_ENDPOINT;
 pub use subnet::Subnet;
 
 pub const JSON_RPC_VERSION: &str = "2.0";
-pub const IPC_GATEWAY_ADDR: u64 = 64;
+pub const DEFAULT_IPC_GATEWAY_ADDR: u64 = 64;
```
adlrocha, line 28: IIRC Henrique is usin this somewhere also, we may need to propagate the change there also. Thanks.

src/server/handlers/mod.rs
```diff
@@ -41,15 +44,16 @@ impl Handlers {
...
+        let config_handler = HandlerWrapper::ReloadConfig(ReloadConfigHandler::new(
+            config.clone(),
             config_path_string,
         ));
-        handlers.insert(String::from("reload_config"), reload_config);
+        handlers.insert(String::from("reload_config"), config_handler);
+
+        let pool = Arc::new(SubnetManagerPool::from_reload_config(config));
+        let create_subnet = HandlerWrapper::CreateSubnet(CreateSubnetHandler::new(pool));
+        handlers.insert(String::from("create_subnet"), create_subnet);
```
adlrocha, line 56: Same here for the RPC method name.

src/server/handlers/subnet.rs
```diff
@@ -0,0 +1,114 @@
...
+        let conn = RwLockReadGuard::map(connections, |connections| {
+            connections.get(subnet_str).unwrap()
+        });
+
+        Some(conn)
+    }
+}
+
+/// TO BE REPLACED AS A LIB CALL ONCE OTHER PR MERGED
+fn lotus_from_subnet(subnet: &Subnet) -> LotusJsonRPCClient<JsonRpcClientImpl> {
```
adlrocha, line 104: Remember to use Henrique's implementation.

adlrocha (approved):
Now sis simpler. Thanks. LGTM.

PR #73, opened 2023-03-06
IPC-68: Max peers per query

### LINKED ISSUE #68: IPLD Resolver: Limit Bitswap peer count
https://github.com/consensus-shipyard/ipc-agent/pull/55 implemented a strategy to first contact the connected peers during Bitswap sync, then go to all other known but not-currently-connection peers. 

Assuming there could be hundreds or thousands of agents known, it would be better to always limit the number of outgoing connections by only contacting so many at any time. 

Add a configuration value to contact N peers at a time, starting with the ones already connected, but shuffle them first, and take some of the known ones if there are less than N connected who can serve a subnet. If the lookup fails, take the next N from the known peers. Repeat until we run out of fallbacks.

aakoshh:
Closes #68 

Changes the strategy for how many peers to contact in bitswap, from "first to all connected, then if failed then all other known providers" to "the first N of connected and known, then the next N, and so on". Also shuffles the peers so we don't always go to the same one.

### DIFF
--- a/ipld/resolver/src/service.rs
+++ b/ipld/resolver/src/service.rs
@@ -20,6 +20,7 @@ use libp2p::{
 use libp2p::{identify, ping};
 use libp2p_bitswap::BitswapStore;
 use log::{debug, error, trace, warn};
+use rand::seq::SliceRandom;
 use tokio::select;
 use tokio::sync::oneshot::{self, Sender};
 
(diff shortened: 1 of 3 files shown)


ipld/resolver/src/service.rs
```diff
@@ -322,20 +334,78 @@ impl<P: StoreParams> Service<P> {
...
+        if peers.is_empty() {
+            send_resolve_result(response_channel, Err(anyhow!(NoKnownPeers(subnet_id))));
+        } else {
+            // Connect to them in a random order, so as not to overwhelm any specific peer.
+            peers.shuffle(&mut rand::thread_rng());
+
+            // Prioritize peers we already have an established connection with.
+            let (connected, known) = peers
+                .into_iter()
+                .partition::<Vec<_>, _>(|id| self.swarm.is_connected(id));
```
adlrocha, line 385: sanity-check: partitioning maintains the order, right? If not we should probably shuffle after this.

adlrocha (approved):
Nice feature. LGTM.

ipld/resolver/src/service.rs
aakoshh: Great question! It does, it just loops through the input and appends to the two outputs. The unstable `partition_in_place` variant doesn't maintain the order. 

I actually started with shuffling both after, then concatenating, but didn't like to see all that code. So I thought: if we have a random list, and we partition it in a way that doesn't maintain the partial order of items, it's still shuffled, right?

PR #74, opened 2023-03-06
IPC-70: Bootstrap Eclipse attack protection 

### LINKED ISSUE #70: IPLD Resolver: Eclipse attack protection
#64 added the mechanism of adding the addresses from `Identify` events to the k-table, otherwise incoming connections (presumably from ephemeral addresses) to Kademlia were treated as unroutable. 

The problem with that solution is that a group of malicious nodes can quickly identify themselves to the node to stuff its k-table full of working but adversarial addresses, and crowd out legitimate peers, eclipsing the node. 

The solution is to make sure we bootstrap first and add peers which we discovered on our own, before we accept self-identified records. The identify only runs once every 5 minutes by default, so we should cache a limited number of them until the bootstrapping is over. 

See https://eprint.iacr.org/2018/236

aakoshh:
Closes #70 

Changes the `Discovery` behaviour to buffer `Identify::Info` records while the Kademlia bootstrapping is ongoing.

### DIFF
--- a/ipld/resolver/src/behaviour/discovery.rs
+++ b/ipld/resolver/src/behaviour/discovery.rs
@@ -14,7 +14,7 @@ use libp2p::{
     identify::Info,
     kad::{
         handler::KademliaHandlerProto, store::MemoryStore, InboundRequest, Kademlia,
-        KademliaConfig, KademliaEvent, KademliaStoreInserts, QueryId,
+        KademliaConfig, KademliaEvent, KademliaStoreInserts, QueryId, QueryResult,
     },
     multiaddr::Protocol,
     swarm::{
@@ -87,6 +87,8 @@ pub struct Behaviour {
     target_connections: usize,
     /// Interval between random lookups.
     lookup_interval: Interval,
+    /// Buffer incoming identify requests until we have finished the bootstrap.
+    bootstrap_buffer: Option<Vec<(PeerId, Info)>>,
     /// Events to return when polled.
     outbox: VecDeque<Event>,
 }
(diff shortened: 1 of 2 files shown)


ipld/resolver/src/behaviour/discovery.rs
```diff
@@ -177,10 +184,20 @@ impl Behaviour {
     /// This seems to be the only way, because Kademlia rightfully treats
     /// incoming connections as ephemeral addresses, but doesn't have an
     /// alternative exchange mechanism.
-    pub fn add_identified(&mut self, peer_id: &PeerId, info: &Info) {
+    pub fn add_identified(&mut self, peer_id: &PeerId, info: Info) {
         if info.protocols.contains(&self.protocol_name) {
-            for addr in info.listen_addrs.iter().cloned() {
-                self.add_address(peer_id, addr);
+            // If we are still in the process of bootstrapping peers, buffer the incoming self-identify records,
```
adlrocha, line 189: How this solve the problem? Even if we block during the bootstrap stage, an attacker could create a lot of sybils an try to connect to me to eventually evict the honest nodes from my routing tables. Or am I missing something?

adlrocha (approved):
The implementation LGTM, I just don't see how strong the protection is. This doesn't impose a huge overhead and overall make sense so we can keep it either way.

ipld/resolver/src/behaviour/discovery.rs
aakoshh: The paper linked to the issue is a good read, and it contains more countermeasures than this, such as not accepting too many IP addresses from the same subnet range. 

Actually, I have no idea if the current Kademlia protects against multiple peer IDs listening on the same address.

The issue here is how the k-table works. Say you put off bootstrapping for a while, and you get a huge influx of connections. The k-table has 256 buckets, one for each distance, with 20 places each; once those are filled, it will only put in more peers if someone from the 20 in the bucket becomes inresponsive. So, an attacker could send lots of connections it controls to take up as many of these buckets as possible, and by the time you have bootstrapped, the nodes you found would only be eligible if the attacker stops responding to pings. 

Because I think we work on `PeerId`, a hash of a public key, it's difficult to target a specific other peer to systematically fill up the k-table, but with a huge amount of random peer IDs, it should fill up eventually.

By not accepting incoming records until the bootstrap is over, we prioritise addresses we find on our own, rather than ones pushed upon us, so it should be a bit more trustworthy (as long as the bootstrap nodes are honest).

PR #75, opened 2023-03-06
IPC-67: Gossip to new joiners

### LINKED ISSUE #67: IPLD Resolver: Gossip to new joiners
I had an idea for how to bring newcomers up to speed quicker than having to wait for the periodic republish in https://github.com/consensus-shipyard/ipc-agent/pull/55#discussion_r1125478850

The idea is that when a peer first receives a record from another one, it could send out its own record to make sure that peer has it as well, unless it has recently done so already.

aakoshh:
Closes #67 

So far we published provider records whenever they changed, and independently at regular intervals. This presented a trade-off in that new joiners would have to wait until they get a full picture of subnet providers. In the #64 I had to set the publish interval low, because currently records are only accepted once the listening address of the other party is known, and it takes time for `Identify` and `Kademlia` to work. 

The PR changes this so that records are also published at the following triggers:
* A peer subscribes to the membership topic - this indicates we just established a Gossipsub connection with them and they might be interested to know about our subnets
* A peer became routable - this indicates that we just received their address through Identify or Kademlia; it's possible that they already subscribed in Gossipsub, but we did not know their listening address at the time
* We received the first provider record from a peer - unlike the first point and Identify, this doesn't need for us to be connected to them; but they are new (to us) and they might be interested in what we provide

To protect against rapid republishing, there's a new `min_time_between_publish` setting and we will not publish if:
* the last publish time was no longer than `min_time_between_publish` ago
* the next publish time is no longer than `min_time_between_publish` in the future
. 
The second condition is for a technical reason: in the smoke test if I published immediately, the test failed, so when we get the subscription, we did not have Identify yet; therefore I delayed the publishing by `min_time_betwen_publish`. But then I want to protect against peers constantly triggering this behaviour and pushing out publishing indefinitely.

The `min_time_between_publish` setting should be set around the network propagation time.

### DIFF
--- a/ipld/resolver/src/behaviour/membership.rs
+++ b/ipld/resolver/src/behaviour/membership.rs
@@ -286,6 +286,16 @@ impl Behaviour {
 
     /// Publish our provider record when we encounter a new peer, unless we have recently done so.
     fn publish_for_new_peer(&mut self, peer_id: PeerId) {
+        if self.subnet_ids.is_empty() {
+            // We have nothing, so there's no need for them to know this ASAP.
+            // The reason we shouldn't disable periodic publishing of empty
+            // records completely is because it would also remove one of
+            // triggers for non-connected peers to eagerly publish their
+            // subnets when they see our empty records. Plus they could
+            // be good to show on metrics, to have a single source of
+            // the cluster size available on any node.
+            return;
+        }
         let now = Timestamp::now();
         if self.last_publish_timestamp > now - self.min_time_between_publish {
             debug!("recently published, not publishing again for peer {peer_id}");
--- a/ipld/resolver/src/behaviour/membership.rs
+++ b/ipld/resolver/src/behaviour/membership.rs
@@ -210,7 +210,7 @@ impl Behaviour {
     /// Call this method when the discovery service learns the address of a peer.
     pub fn set_routable(&mut self, peer_id: PeerId) {
         self.provider_cache.set_routable(peer_id);
-        self.publish_to_new_peer(peer_id);
+        self.publish_for_new_peer(peer_id);
     }
 
     /// Mark a peer as unroutable in the cache.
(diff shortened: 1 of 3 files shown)


ipld/resolver/src/behaviour/membership.rs
```diff
@@ -188,14 +199,18 @@ impl Behaviour {
...
         Ok(())
     }
 
     /// Mark a peer as routable in the cache.
     ///
     /// Call this method when the discovery service learns the address of a peer.
     pub fn set_routable(&mut self, peer_id: PeerId) {
-        self.provider_cache.set_routable(peer_id)
+        self.provider_cache.set_routable(peer_id);
+        self.publish_for_new_peer(peer_id);
```
adlrocha, line 213: This assumes that when a new peer joins the topic all peers in their topic will share their membership, right? This will lead to bursts of messages with ever new joiner. I think this is fine with the time safe-guard you added, but we'll have to be careful as the size of the network increases as this may be too much and we may have to prune in some way who broadcasts their membership triggered by new joiners to spread the membership updates.

ipld/resolver/src/behaviour/membership.rs
```diff
@@ -233,9 +244,72 @@ impl Behaviour {
...
+            // triggers for non-connected peers to eagerly publish their
+            // subnets when they see our empty records. Plus they could
+            // be good to show on metrics, to have a single source of
+            // the cluster size available on any node.
+            return;
+        }
+        let now = Timestamp::now();
+        if self.last_publish_timestamp > now - self.min_time_between_publish {
+            debug!("recently published, not publishing again for peer {peer_id}");
+        } else if self.next_publish_timestamp <= now + self.min_time_between_publish {
```
adlrocha, line 302: I like this check!

adlrocha (approved):
LGTM. Amazong work!

ipld/resolver/src/behaviour/membership.rs
aakoshh: You are completely right, it does lead to annoying bursts. Maybe they could choose some random time to publish, based on the cluster size. 

I just thought about an alternative, but it would need some rework: Instead of everyone publishing their own when they learn about a new guy, we could leave republishing only to the ones the new guy subscribes with. We don't have a point-to-point protocol here, but if we cached the incoming records as they are, we could publish them as they are, and I think the Gossipsub duplicate filter could make sure that everyone other than our new joiner would ignore the message. Does that sound plausible?

adlrocha: > I think the Gossipsub duplicate filter could make sure that everyone other than our new joiner would ignore the message

Assuming this is in place (which I guess in Rust is also the case), I think this would work with a small caveat. Re-using messages may lead to them not being forwarded due to the filter, and not being propagated enough to reach the new joiner. It is probably best if we leave it is as-is for now and maybe think of an optimization if this becomes an issue, WDYT?

PR #76, opened 2023-03-07
Query validator set in Json Rpc server

cryptoAtwill:
Support querying validator set in IPC-Agent Json Rpc server.

@dnkolegov please help check if this is aligned with your requirements.


src/server/handlers/validator.rs
```diff
@@ -0,0 +1,59 @@
...
+#[async_trait]
+impl JsonRPCRequestHandler for QueryValidatorSetHandler {
+    type Request = QueryValidatorSetParams;
+    type Response = QueryValidatorSetResponse;
+
+    async fn handle(&self, request: Self::Request) -> anyhow::Result<Self::Response> {
+        let tip_set = Cid::from_str(&request.tip_set)?;
+
+        let config = self.config.get_config();
+        let subnet = match config.subnets.get(&request.subnet) {
```
adlrocha, line 47: You pass as an argument the target subnet id to get the validator set from, and this actor lives in the parent, so here you should be getting the parent.

src/server/handlers/validator.rs
```diff
@@ -0,0 +1,59 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct QueryValidatorSetParams {
+    pub subnet: String,
+    pub tip_set: String,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct QueryValidatorSetResponse {
+    /// The address of the created subnet
+    pub validator_set: Vec<Validator>,
```
adlrocha, line 24: I will tag you on a PR that I am working on as this may change slightly.

adlrocha (approved):
This is perfect in terms of requirements. Make the CI happy and wait for merging to the change we need to introduce in the subnet actor so you can conveniently test it. Thanks.

dnkolegov (approved):
What is the full JSONRPC method name? 

It would be great to add a dumb test to check that the JSON RPC client can get a validator set from the server. I would also use that test to test that my client is compatible with your server

src/server/handlers/validator.rs
cryptoAtwill: np np, will keep an eye for this

cryptoAtwill:
@dnkolegov Sure sure. We are introducing quite some changes and I have to run another round of tests. I will send you the test script or curl command for you to try.

dnkolegov:
> @dnkolegov Sure sure. We are introducing quite some changes and I have to run another round of tests. I will send you the test script or curl command for you to try.

That's not for `curl`, that's for Eudico https://github.com/consensus-shipyard/lotus/tree/onchain-reconfiguration/chain/consensus/mir/rpc.

src/server/handlers/validator.rs
```diff
@@ -0,0 +1,67 @@
...
```
adlrocha, line 27: We should introduce the new `ValidatorSet`struuct introduced in this PR: https://github.com/consensus-shipyard/ipc-actors/pull/61. We need all of it for reconfiguration.

adlrocha (changes requested):
Some minor updates before merge.

src/server/handlers/validator.rs
cryptoAtwill: @adlrocha I have updated the types, please take a look.

adlrocha (approved):
LGTM. Thanks!

PR #77, opened 2023-03-07
update config cli

cryptoAtwill:


src/cli/commands/mod.rs
```diff
@@ -75,6 +80,7 @@ pub async fn cli() {
     let r = match &args.command {
         Commands::Daemon(args) => LaunchDaemon::handle(global, args).await,
         Commands::CreateSubnet(args) => CreateSubnet::handle(global, args).await,
+        Commands::ReloadConfig(args) => ReloadConfig::handle(global, args).await,
```
adlrocha, line 83: Same here for the grouping.

src/jsonrpc/mod.rs
```diff
@@ -119,12 +119,18 @@ struct JsonRpcResponse<T> {
...
     fn from(j: JsonRpcResponse<T>) -> Self {
         if j.error.is_some() {
-            Err(anyhow!("json_rpc error: {:}", j.error.unwrap()))
-        } else {
+            return Err(anyhow!("json_rpc error: {:}", j.error.unwrap()));
+        }
+        if j.result.is_some() {
             Ok(j.result.unwrap())
+        } else {
+            // the result is not found, but it is possible T could (),  means the
```
adlrocha, line 130: > but it is possible T could ()

The comment is great but it is not clear what this means, do you mind rewriting it to make it clearer? Thanks.

src/cli/commands/config.rs
```diff
@@ -0,0 +1,56 @@
...
+
+#[async_trait]
+impl CommandLineHandler for ReloadConfig {
+    type Arguments = ReloadConfigArgs;
+
+    async fn handle(global: &GlobalArguments, arguments: &Self::Arguments) -> anyhow::Result<()> {
+        log::debug!("reload config with args: {:?}", arguments);
+
+        let params = ReloadConfigParams {
+            path: arguments.path.clone(),
```
adlrocha, line 26: I am wondering if we should pass the full config instead of a path to the json rpc method for config reload, what if the cli and the jsonrpc server live in different machines?

adlrocha (approved):
LGTM, a few comments I want your opinion about, but this is mergeable. Thanks.

src/cli/commands/config.rs
cryptoAtwill: I could be wrong, but I have this feeling that config reload is a really privileged operation, we should put some restrictions as not to expose this method to the network, who know what ppl might do. Probably only same machine invocation has fewer risks? If we wan to pass the Config struct, it's setting the configs and we should persists those updated config to survive server restarts, probably it's not called a config reload and we can handle it in another method?

src/jsonrpc/mod.rs
cryptoAtwill: I have updated, hope it's clearer.

src/cli/commands/mod.rs
```diff
@@ -22,6 +26,7 @@ enum Commands {
     /// due to the convention from `lotus` and the expected behavior from the filecoin user group.
     Daemon(LaunchDaemonArgs),
     CreateSubnet(CreateSubnetArgs),
+    ReloadConfig(ReloadConfigArgs),
```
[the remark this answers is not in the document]
adlrocha: Tracking a related issue here: https://github.com/consensus-shipyard/ipc-agent/issues/87

src/cli/commands/config.rs
adlrocha: >  If we wan to pass the Config struct, it's setting the configs and we should persists those updated config to survive server restarts, probably it's not called a config reload and we can handle it in another method?

This is a good point. 

> I could be wrong, but I have this feeling that config reload is a really privileged operation, we should put some restrictions as not to expose this method to the network, who know what ppl might d

I am not that worried about this at this point because we need to figure out authentication and access rights. For now we are assuming that for M2 users won't expose their agent publicly. 

I guess for simplicity we can leave it like this for now, but do you mind tracking this in an issue so we tackle it in the future?

src/jsonrpc/mod.rs
```diff
@@ -119,12 +119,18 @@ struct JsonRpcResponse<T> {
...
     fn from(j: JsonRpcResponse<T>) -> Self {
         if j.error.is_some() {
-            Err(anyhow!("json_rpc error: {:}", j.error.unwrap()))
-        } else {
+            return Err(anyhow!("json_rpc error: {:}", j.error.unwrap()));
+        }
+        if j.result.is_some() {
             Ok(j.result.unwrap())
+        } else {
+            // The result is not found, but it is possible T could the rust unit type: (), i.e. the
```
adlrocha, line 130:
```suggestion
            // The result is not found, but it is possible T could be the rust unit type: (), i.e. the
```

adlrocha (approved):
LGTM, let's track the comment about passing the full configuration to the agent for the future. We won't need it now, but in the future I expect users to interact with a remote agent and we may need to support that... we'll see.

src/jsonrpc/mod.rs
adlrocha: Fixed. Thanks.

src/cli/commands/config.rs
cryptoAtwill: Added, see #88.

PR #81, opened 2023-03-07
Add join, leave and kill subnet method

cryptoAtwill:
Add `join`, `leave` and `kill` subnet json rpc handlers.


src/server/handlers/join.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright 2022-2023 Protocol Labs
```
adlrocha, line 1: I think we should group handles by topical modules. In this case, add, join, kill and leave should be part of a subnet or manager module, WDYT?

src/server/handlers/join.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+#[async_trait]
+impl JsonRPCRequestHandler for JoinSubnetHandler {
+    type Request = JoinSubnetParams;
+    type Response = ();
+
+    async fn handle(&self, request: Self::Request) -> anyhow::Result<Self::Response> {
+        let conn = match self.pool.get(&request.subnet) {
```
adlrocha, line 43: This is a bug, you are getting the connection to the subnet ID, not to the parent, and you want to interact with the subnet actor in the parent.

src/server/handlers/kill.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+}
+
+#[async_trait]
+impl JsonRPCRequestHandler for KillSubnetHandler {
+    type Request = KillSubnetParams;
+    type Response = ();
+
+    async fn handle(&self, request: Self::Request) -> anyhow::Result<Self::Response> {
+        let conn = match self.pool.get(&request.subnet) {
```
adlrocha, line 39: Same as above, you should get the conn for the parent of the subnet here.

src/server/handlers/leave.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+}
+
+#[async_trait]
+impl JsonRPCRequestHandler for LeaveSubnetHandler {
+    type Request = LeaveSubnetParams;
+    type Response = ();
+
+    async fn handle(&self, request: Self::Request) -> anyhow::Result<Self::Response> {
+        let conn = match self.pool.get(&request.subnet) {
```
adlrocha, line 39: Same as above, you should get the conn for the parent of the subnet here.

src/server/handlers/mod.rs
```diff
@@ -53,29 +100,25 @@ impl Handlers {
...
         Ok(Self { handlers })
     }
-
-    pub async fn handle(&self, method: Method, params: Value) -> Result<Value> {
-        if let Some(wrapper) = self.handlers.get(&method) {
-            match wrapper {
-                HandlerWrapper::CreateSubnet(handler) => {
-                    let r = handler.handle(serde_json::from_value(params)?).await?;
-                    Ok(serde_json::to_value(r)?)
-                }
```
adlrocha, line 71: I am wondering if the body for each of this branches can be de-duplicated without the need of a macro, it is the same for all of them

src/server/handlers/mod.rs
```diff
@@ -18,19 +24,60 @@ use serde_json::Value;
 use std::collections::HashMap;
 use std::sync::Arc;
 
-pub type Method = String;
+/// A util macro to create the handlers and do the method routing.
+/// This macro generates the `handle` method. In the method, it just matches the handler with
+/// different enum types to get the inner handler instance. Every time a new handler is added,
+/// a new match statement must be copy-pasted. With this macro, we can reduce a lot of repetitive
+/// copy paste.
+macro_rules! route_handlers {
```
adlrocha, line 32: Would you mind keeping this PR only for the implementation of the handlers and move the macro proposal to another PR? It may potentially require a longer discussion and I don't want to block these commands from being merged. Thanks!

adlrocha (changes requested):
Amazing work. A few bugs, and a proposal to move the macro to another PR to avoid delaying these commands due to a discussion over the macro. Thanks.

src/server/handlers/mod.rs
cryptoAtwill: sure sure

src/server/handlers/leave.rs
cryptoAtwill: updated!

src/server/handlers/kill.rs
cryptoAtwill: updated!

src/server/handlers/join.rs
cryptoAtwill: updated!

src/server/handlers/join.rs
cryptoAtwill: updated and placed in a separate mod.

src/server/handlers/mod.rs
cryptoAtwill: No need enum and macro now, I found a way to handle it already!

src/server/handlers/join.rs
```diff
@@ -40,15 +40,19 @@ impl JsonRPCRequestHandler for JoinSubnetHandler {
     type Response = ();
 
     async fn handle(&self, request: Self::Request) -> anyhow::Result<Self::Response> {
-        let conn = match self.pool.get(&request.subnet) {
+        let subnet = SubnetID::from_str(&request.subnet)?;
+        let parent = subnet
+            .parent()
+            .ok_or_else(|| anyhow!("no parent found"))?
+            .to_string();
+        let conn = match self.pool.get(&parent) {
```
adlrocha: Wouldn't it make more sense for `pool.get` to accept a `SubnetID` as an argument instead of a string? It would remove the need of an additional check inside of get and having to do a lot of back and forth between types.

adlrocha (approved):
LGTM. Just one last comment regarding using `SubnetID` as the argument type to `pool.get`. Thanks.

src/server/handlers/join.rs
cryptoAtwill: yeah, make sense, haha. Update this now.

PR #82, opened 2023-03-07
IPC-69: IPLD Resolver documentation

### LINKED ISSUE #69: IPLD Resolver: Documentation
Write a README and provide sequence diagrams for the Resolver to make it easier to understand how it works and how to use it.

aakoshh:
Closes #69 

The PR creates a [docs](https://github.com/consensus-shipyard/ipc-agent/tree/ipc-69-docs/docs) directory with a README that contains illustrative diagrams to show the common usecase of checkpointing and then the internal mechanisms of the IPLD Resolver. See the link for a rendered view.

Currently it's just a single page; I'm sure it would make sense to break it up into sub-pages very soon if we keep adding to it. 

The diagrams can be produced by `make diagrams`. CI also checks that there are no uncommitted changes, and there's an example git hook to make it happen on pre-commit.


docs/README.md
```diff
@@ -0,0 +1,89 @@
...
+Please refer to the [IPD Agent Design](https://docs.google.com/document/d/14lkRRv6MQYnuEfp2GoGngdD8Q5YgfE38D8HTZWKgKf4) document for details on the agent.
+
+
+# IPLD Resolver
+
+The [IPLD Resolver](../ipld/resolver) is a library that IPC Agents can use to exchange data between subnets in IPLD format.
+
+## Checkpointing
+
+The most typical use case would be the propagation of checkpoints from child subnets to the parent subnet.
```
adlrocha, line 14: For M2 this is not quite the case, the resolver will be used to fetch cross-net messages.

docs/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Checkpoint Schema
+
+One possible conceptual model of checkpointing is depicted by the following Entity Relationship diagram:
+
+![Checkpoint Schema](diagrams/checkpoint_schema.png)
+
+It shows that the Subnet Actor in the parent subnet governs the power of validators in the child subnet by proposing _Configurations_, which the child subnet is free to adopt in its _Epochs_ when the time is right, communicating back the next adopted config via _Checkpoints_.
```
adlrocha, line 22: > communicating back the next adopted config via _Checkpoints_.

Note, we don't have a feedback loop yet from child subnet to parents in place yet.

docs/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+Once enough signatures are collected to form a Quorum Certificate over the checkpoint (the specific rules are in the jurisdiction of the Subnet Actor), the checkpoint is submitted to the parent ledger.
+
+However, the submitted checkpoint does not contain the raw messages, only the meta-data. The content needs to be resolved using the IPC Resolver, as indicated by the dotted line.
+
+### Checkpoint Submission and Resolution
+
+The following sequence diagram shows one possible way how checkpoints can be submitted from the child to the parent subnet.
+
+It depicts two validators: one only participating on the parent subnet, and the other on the child subnet; the latter has to also run at least a full node on the parent subnet. Both validators run one IPC AGent each.
```
adlrocha, line 34:
```suggestion
It depicts two validators: one only participating on the parent subnet, and the other on the child subnet; the latter has to also run at least a full node on the parent subnet. Both validators run one IPC Agent each.
```

docs/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+The diagram shows that at the end of the epoch the child subnet validators produce a Quorum Certificate over the checkpoint, which some of their agents submit to the parent subnet.
+
+After that, the parent subnet nodes reach out to their associated IPC Agent to resolve the checkpoint, which the Agent does by communicating with some of its child-subnet peers.
```
adlrocha, line 38: In the current implementation the raw checkpoint is submitted to the parent and persisted in the state so there's no need to resolve it from the child.

adlrocha (approved):
Left a few clarifications, but otherwise it is perfect. Thanks.

docs/README.md
aakoshh: I may be misunderstanding what that means. I thought a cross-net message is a catch-all phrase for bottom-up and top-down messages. I understand that strictly they probably mean this:

```
/root |- /sub1  >
      |          |
      |          |
      |     cross-msg      
      |          |
      |          |
      |- /sub2  < 
```
But I thought that would be a combination of a bottom-up and a top-down message, and involve a checkpoint.

docs/README.md
aakoshh: Okay no problem, just wanted to orient the reader.

docs/README.md
aakoshh: Yes, I will change this to "checkpoint contents".

PR #84, opened 2023-03-08
IPC-65: Metrics

### LINKED ISSUE #65: IPLD Resolver: Prometheus metrics
Maintain Prometheus metrics, similar to how the embedded Bitswap library is doing, so that we can have track how many subnets we know about, how many providers we skip, etc.

aakoshh:
Closes #65 

The PR adds a bunch of Prometheus metrics to the library, and a `Service::register_metrics` method, similar to the one `Bitswap` has.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -217,9 +217,9 @@ dependencies = [
 
 [[package]]
 name = "asn1-rs"
-version = "0.5.1"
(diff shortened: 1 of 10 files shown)


adlrocha (approved):
LGTM. This is going to be really useful in the future. 🙏

PR #89, opened 2023-03-09
list subnets

cryptoAtwill:
List subnets in a gateway actor.

### DIFF
--- a/src/server/handlers/manager/list.rs
+++ b/src/server/handlers/manager/list.rs
@@ -1,0 +1,46 @@
+// Copyright 2022-2023 Protocol Labs
+// SPDX-License-Identifier: MIT
+//! List subnets in gateway actor
+
+use crate::manager::{SubnetInfo, SubnetManager};
+use crate::server::handlers::manager::subnet::SubnetManagerPool;
+use crate::server::JsonRPCRequestHandler;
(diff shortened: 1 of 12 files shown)


src/server/handlers/manager/list.rs
```diff
@@ -0,0 +1,46 @@
+// Copyright 2022-2023 Protocol Labs
```
adlrocha, line 1: Should we call this file maybe `list_subnets.rs`?

src/server/handlers/manager/list.rs
```diff
@@ -0,0 +1,46 @@
...
+use async_trait::async_trait;
+use ipc_sdk::subnet_id::SubnetID;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::str::FromStr;
+use std::sync::Arc;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ListSubnetsParams {
+    pub subnet: String,
```
adlrocha, line 18: As mentioned above, unfortunately we may have to pass the gateway address here.

adlrocha (approved):
Some slight bugs, let's fix them, try it in Lotus, and we are good to go.

adlrocha (approved):

PR #90, opened 2023-03-09
IPC-86: Rate limit

### LINKED ISSUE #86: IPLD Resolver: Rate limiting in Bitswap
I can't remember seeing any policy in the `Bitswap` behaviour to limit the amount of data any particular peer can request. We should make sure nobody can abuse the openness of the protocol by downloading unreasonable amounts of data.

The best option seems to be to inspect the `BitswapResponse`s sent back by the `Bitswap` to its handler as it passes through the `content::Behaviour`. Here we have access to `NotifyHandler::peer_id` to maintain some stats about how much data we have served to any given peer in the latest time window. Then, we can look at incoming requests in `content::Behaviour::on_connection_handler_event` and deny them if they violate the policy.

aakoshh:
Closes #86 

The PR uses the [gcra](https://crates.io/crates/gcra) library to rate limit the amount of data that can go to any individual `MultiAddress` via `Bitswap`. The `content::Config` gets two new fields, `rate_limit_bytes` and `rate_limit_period`, so that we can configure e.g. that any peer cannot download more than 10MB/hour from us. 

The `Client` has a new `update_rate_limit` method we can use to adjust this value on the fly, if we see that this is going to be too tight. The `gcra` library was picked because it allows passing in the limit with each request, unlike for example the [governor](https://crates.io/crates/governor). 

To make it work, I had to fork `libp2p-bitswap`, because normally it sends the response straight to the connection handler, which writes it to the socket. To inspect the response, I had to redirect the channels, which required the hitherto private `BitswapResponse` type to be public. 

I see that `libp2p-bitswap` has metrics like `THROTTLED_INBOUND`, however, they aren't used. This suggests that rate limiting was supposed to be part of the library, but apart from https://github.com/ipfs-rust/libp2p-bitswap/issues/22 saying it's a bit naive at the moment, I didn't find anything on this. I thought making the aforementioned type public is the lightest touch. 

Another option which would have resulted in much less code and indirection on our part would have been to extend the `BitswapEvent` to not only report progress on our own queries, but the responses sent. I didn't want to break the API for anyone who uses this library, without talking to the devs, though. Maybe they would prefer to add throttling directly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -72,7 +72,7 @@ version = "0.7.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9e8b47f52ea9bae42228d07ec09eb676433d7c4ed1ebdf0f1d1c29ed446f1ab8"
 dependencies = [
- "cfg-if",
+ "cfg-if 1.0.0",
  "cipher 0.3.0",
  "cpufeatures",
  "opaque-debug",
@@ -84,7 +84,7 @@ version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "433cfd6710c9986c576a25ca913c39d66a6474107b406f34f91d4a8923395241"
 dependencies = [
- "cfg-if",
+ "cfg-if 1.0.0",
  "cipher 0.4.3",
  "cpufeatures",
 ]
@@ -367,7 +367,7 @@ checksum = "233d376d6d185f2a3093e58f283f60f880315b6c60075b01f36b3b85154564ca"
 dependencies = [
  "addr2line 0.19.0",
  "cc",
- "cfg-if",
+ "cfg-if 1.0.0",
  "libc",
  "miniz_oxide",
  "object 0.30.3",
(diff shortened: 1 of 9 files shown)


adlrocha (approved):
While I agree that this is a really useful feature, I am wondering if this may be an early optimization at this point. In the context of IPC, I don't think there will be useful content or an incentive to free-ride agents (I may be wrong). 

As long as it doesn't break anything I am fine introducing it now (we can always keep the rate limiter to 0 initially for IPC). But as mentioned in the comments, my main concern is the increased complexity and the need to rely on a fork.

aakoshh:
I agree that a fork is not ideal, I opened a PR to the upstream repo. 

I also agree that most likely nobody is going to abuse our agents. However, library is agnostic of what you want to download, so someone can totally use it to download the entire state tree if they ask for that CID. At least in Fendermint I don’t want to go into gymnastics to filter what exactly can be downloaded, so this level of fair use limit can be a simple deterrent for people who don’t want to risk being unable to download stuff that matters, or to not latch on a single peer too much.

PR #95, opened 2023-03-10
Update lotus api

cryptoAtwill:
Update lotus get subnet actor state to work with lotus node.

### DIFF
--- a/src/lotus/client.rs
+++ b/src/lotus/client.rs
@@ -303,6 +303,8 @@ impl<T: JsonRpcClient + Send + Sync> LotusClient for LotusJsonRPCClient<T> {
             },
             [CIDMap::from(tip_set)]]
         );
+        log::debug!("sending {params:?}");
+
         let r = self
             .client
             .request::<IPCReadSubnetActorStateResponse>(
--- a/src/lotus/client.rs
+++ b/src/lotus/client.rs
@@ -51,6 +51,12 @@ mod methods {
 
 /// The default gateway actor address
 const GATEWAY_ACTOR_ADDRESS: &str = "f064";
+/// The default state wait confidence value
+const STATE_WAIT_CONFIDENCE: u8 = 5;
+/// We dont set a limit on the look back epoch, i.e. check against latest block
+const STATE_WAIT_LOOK_BACK_NO_LIMIT: i8 = -1;
+/// We are not replacing any previous messages
+const STATE_WAIT_ALLOW_REPLACE: bool = false;
 
 /// The struct implementation for Lotus Client API. It allows for multiple different trait
 /// extension.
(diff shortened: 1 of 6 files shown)


src/lotus/client.rs
```diff
@@ -51,6 +51,12 @@ mod methods {
 
 /// The default gateway actor address
 const GATEWAY_ACTOR_ADDRESS: &str = "f064";
+/// The default state wait confidence value
+const STATE_WAIT_CONFIDENCE: u8 = 5;
+/// We dont set a limit on the look back epoch, i.e. check against latest block
+const STATE_WAIT_LOOK_BACK_NO_LIMIT: i8 = -1;
+/// We are not replacing any previous messages
+const STATE_WAIT_ALLOW_REPLACE: bool = true;
```
adlrocha, line 59: I am wondering if we should set this to `false`. Having it to `true` means that if the message is replaced by a new one but with different parameters and it is validated, `StateWaitMsg` for this message will also shoe as a success.

It may also be a good idea to update the comment accordingly.

src/lotus/client.rs
```diff
@@ -288,9 +299,23 @@ impl<T: JsonRpcClient + Send + Sync> LotusClient for LotusJsonRPCClient<T> {
...
+        let parent = subnet_id
+            .parent()
+            .ok_or_else(|| anyhow!("no parent found"))?
+            .to_string();
+        let actor = subnet_id.subnet_actor().to_string();
+        let params = json!([
+            {
+                "Parent": parent,
+                "Actor": actor
+            },
```
adlrocha, line 314: Serializing `SubnetID` directly into json doesn't work due to the `serialize_tuple`, right? 

I would suggest adding a `to_json` method in `SubnetID` that does this so we don't have to do it every time, or include in the agent a `subnet_id_to_json` function to avoid having to remember how to do it like here.

src/lotus/message/ipc.rs
```diff
@@ -29,3 +28,17 @@ pub struct IPCReadSubnetActorStateResponse {
     pub check_period: ChainEpoch,
     pub validator_set: ValidatorSet,
 }
+
+#[derive(Deserialize, Serialize, Debug)]
```
adlrocha, line 32: You had to duplicate the structs here because there `serialize_tuple` didn't work, right? Also, double-check always the order of the fields when using `serialize_tuple`, as the fields are not the indexed by name but deterministically by their order (their underlying representation is a list)  

Do you mind adding a note here of why you need to re-write these structs?

src/lotus/client.rs
adlrocha: I see that we only make this serialization here, if this is the case maybe we can keep it as-is for now.

adlrocha (approved):
LGTM. Thanks. Double-check that you changed all the relevant methods that have changed in Lotus before merging (I only see the `read_subnet_state` but I don't know if we have implemented any other yet.

src/lotus/client.rs
cryptoAtwill: yeah, maybe we keep it for now.

src/lotus/client.rs
cryptoAtwill: when I set it to `false`, it gives this error:https://github.com/consensus-shipyard/lotus/blob/35049ce3171087b9e0d64724332fa20cf3776341/chain/stmgr/searchwait.go#L267. Maybe I have done something wrong?

src/lotus/message/ipc.rs
cryptoAtwill: Yeah, comments added. But the idea is in the actor, it is `deserialize_tuple`, but when returned from json rpc endpoints, it's actually `json` struct. It's not working because the agent is interpreting the tuple as json.

PR #97, opened 2023-03-11
Initial version of Docker file with cache

dnkolegov:
This PR introduces a basic Docker file.


Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+COPY . .
+
+RUN --mount=type=cache,target=$RUSTUP_HOME,from=rust,source=$RUSTUP_HOME \
+    --mount=type=cache,target=$CARGO_HOME,from=rust,source=$CARGO_HOME \
+    --mount=type=cache,target=target \
+    cargo install --root output --path .
+
+
+# Main stage
+FROM debian:bullseye-slim
```
adlrocha, line 26: Why debian here?

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+WORKDIR /app
+
+COPY . .
+
+RUN --mount=type=cache,target=$RUSTUP_HOME,from=rust,source=$RUSTUP_HOME \
+    --mount=type=cache,target=$CARGO_HOME,from=rust,source=$CARGO_HOME \
+    --mount=type=cache,target=target \
+    cargo install --root output --path .
+
```
adlrocha, line 23: We should also expose whatever port we use for the IPC endpoint. Here you have a sample template, we will have to handle them in tests so we can interact between images: https://github.com/consensus-shipyard/ipc-agent/blob/main/config/template.toml

adlrocha (approved):
I haven't tried it, if it works let's get this merged and then improve it as we develop the tests.

Dockerfile
dnkolegov: what should be there? alpine? Debian is a standard choice for Rust

Dockerfile
dnkolegov: Fixed.

Dockerfile
adlrocha: Debian is fine for now, we can change it in the future.

PR #99, opened 2023-03-13
Add crossnet interface

cryptoAtwill:
Update subnet manager interface with crossnet functions:
- fund
- release
- propagate
- whitelist_propagator

### DIFF
--- a/src/manager/lotus.rs
+++ b/src/manager/lotus.rs
@@ -161,7 +161,7 @@ impl<T: JsonRpcClient + Send + Sync> SubnetManager for LotusSubnetManager<T> {
     async fn whitelist_propagator(
         &self,
         _subnet: SubnetID,
-        _postbox_cid: Cid,
+        _postbox_msg_cid: Cid,
         _from: Address,
         _to_add: Vec<Address>,
     ) -> Result<()> {
--- a/src/manager/lotus.rs
+++ b/src/manager/lotus.rs
@@ -154,7 +154,7 @@ impl<T: JsonRpcClient + Send + Sync> SubnetManager for LotusSubnetManager<T> {
         todo!()
     }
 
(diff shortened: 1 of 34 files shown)


src/manager/lotus.rs
```diff
@@ -145,6 +145,27 @@ impl<T: JsonRpcClient + Send + Sync> SubnetManager for LotusSubnetManager<T> {
...
+        todo!()
+    }
+
+    async fn propagate(&self, _subnet: SubnetID, _from: Address, _postbox_cid: Cid) -> Result<()> {
+        todo!()
+    }
+
+    async fn whitelist_propagator(
+        &self,
+        _subnet: SubnetID,
```
adlrocha, line 163: We need to specify the cid of the message to add the propagator to, right?

src/manager/lotus.rs
```diff
@@ -145,6 +145,27 @@ impl<T: JsonRpcClient + Send + Sync> SubnetManager for LotusSubnetManager<T> {
...
+
+    async fn fund(&self, _subnet: SubnetID, _from: Address, _amount: TokenAmount) -> Result<()> {
+        todo!()
+    }
+
+    async fn release(&self, _subnet: SubnetID, _from: Address) -> Result<()> {
+        todo!()
+    }
+
+    async fn propagate(&self, _subnet: SubnetID, _from: Address, _postbox_cid: Cid) -> Result<()> {
```
adlrocha, line 157:
```suggestion
    async fn propagate(&self, _subnet: SubnetID, _from: Address, _postbox_msg_cid: Cid) -> Result<()> {
```

adlrocha (approved):
LGTM, be sure to change the issues I raised before merging. But apart from that let's keep moving :) Thanks.

src/manager/lotus.rs
cryptoAtwill: updated and resolving this.

PR #100, opened 2023-03-13
set network

cryptoAtwill:
Setting the network from env variable. This is because in the config, it's deserializing based on a network, if the network is not set, it will give an error.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,9 +1,21 @@
 // Copyright 2022-2023 Protocol Labs
 // SPDX-License-Identifier: MIT
+use fvm_shared::address::{set_current_network, Network};
 use ipc_agent::cli;
+use num_traits::FromPrimitive;
 
 #[tokio::main]
 async fn main() {
     env_logger::init_from_env(env_logger::Env::new().default_filter_or("info"));
+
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -1,9 +1,21 @@
 // Copyright 2022-2023 Protocol Labs
 // SPDX-License-Identifier: MIT
+use fvm_shared::address::{set_current_network, Network};
 use ipc_agent::cli;
+use num_traits::FromPrimitive;
 
 #[tokio::main]
 async fn main() {
     env_logger::init_from_env(env_logger::Env::new().default_filter_or("info"));
+
```
adlrocha, line 10: My main concern with this is what if one of the subnets is treated as a subnet and others as a mainnet in terms of address type, but I guess we can keep this approach for now and change it if we see that it doesn't work.

adlrocha (approved):
LGTM.

adlrocha (commented):
BTW, remove this from the config file if it is no longer needed: https://github.com/consensus-shipyard/ipc-agent/blob/fc97f950be86e74accd57528afbbfebd56c71579/config/template.toml#L3

PR #104, opened 2023-03-13
Add more cli commands

cryptoAtwill:
Add more cli commands: `join`, `leave` and `kill` subnet.


adlrocha (approved):
Let's not get used to this way of working, but as I am currently testing these let me merge them right away, and I will fix any issue if there are any. Thanks.

PR #107, opened 2023-03-13
New `init-config` cmd to populate empty config in path and several fixes

adlrocha:
This PR: 
- Implements a new `init-config` command that populates an new valid and empty config on a target path. This improves the UX and will come really handy for testing. 
- Adds a new `make build` rule to compile the agent in a `/bin`  output directory (also handy for testing).
- Switches to named parameters for CLI commands.
- Several fixes after testing the subnet manager commands manually.

Next: Testing the actual run of a subnet after joining. //cc @dnkolegov


src/cli/commands/config.rs
```diff
@@ -54,3 +55,38 @@ pub(crate) struct ReloadConfigArgs {
...
+
+    async fn handle(global: &GlobalArguments, _arguments: &Self::Arguments) -> anyhow::Result<()> {
+        let path = global.config_path();
+        log::debug!("initializing empty config file in {}", path);
+
+        let file_path = std::path::Path::new(&path);
+        if let Some(parent) = file_path.parent() {
+            std::fs::create_dir_all(parent)?;
+        }
+        let mut file = std::fs::File::create(&path).map_err(|e| {
```
cryptoAtwill, line 74: we can use tokio::fs, it has async version, which we wont block the thread.

src/config/mod.rs
```diff
@@ -27,6 +27,27 @@ pub use subnet::Subnet;
...
+jsonrpc_api_ws = "wss://example.org/rpc/v0"
+auth_token = "YOUR TOKEN"
+accounts = ["t01"]
+
+[subnets."/root/t01"]
+id = "/root/t01"
+jsonrpc_api_http = "http://127.0.0.1:1235/rpc/v1"
+auth_token = "YOUR TOKEN"
+accounts = ["t01"]
+    "#;
```
cryptoAtwill, line 49: seems there are some extra spaces

cryptoAtwill (approved):

src/config/mod.rs
adlrocha: Good call, fixed.

PR #109, opened 2023-03-14
Add whitelist and propagate methods

cryptoAtwill:
Add whitelist and propagate message push.


src/manager/lotus.rs
```diff
@@ -188,23 +188,57 @@ impl<T: JsonRpcClient + Send + Sync> SubnetManager for LotusSubnetManager<T> {
...
+        )?;
+
+        let message = MpoolPushMessage::new(
+            Address::new_id(DEFAULT_IPC_GATEWAY_ADDR),
+            from,
+            ipc_gateway::Method::Propagate as MethodNum,
+            params.to_vec(),
+        );
+
+        self.mpool_push_and_wait(message).await?;
```
adlrocha, line 210: @cryptoAtwill , something I was wondering here while I was doing tests. Is there any timeout for this await? Or if the message stays pending forever this request stays pending forever? I guess the rpc server has some kind of timeout but I wanted to double-check.

adlrocha (approved):
LGTM, feel free to merge. Thanks!

src/manager/lotus.rs
cryptoAtwill: Currently this wait has no timeout, at least in the agent. We can introduce a timeout, but I guess the question is really what do we do when it times out, we just return an error? But I think it would be a better practice to have a timeout.

PR #110, opened 2023-03-14
fix list subnet to string

cryptoAtwill:
Explicitly convert address to string so that list subnet will work.


adlrocha (approved):
In the end it was an easy one :) I'll fix the Go side of things.

PR #114, opened 2023-03-14
Inject ReloadableConfig into JsonRPCServer

hmoniz:
This injects the ReloadableConfig as a constructor parameter into the JsonRPCServer as opposed to having the JsonRPCServer constructing it. The allows the ReloadableConfig to be resued by other components. In particular, it will allow the CheckpointSubsystem to use it.

### DIFF
--- a/src/manager/checkpoint.rs
+++ b/src/manager/checkpoint.rs
@@ -33,7 +33,7 @@ use crate::lotus::LotusClient;
 const CHAIN_HEAD_REQUEST_PERIOD: Duration = Duration::from_secs(10);
 
 /// The `CheckpointSubsystem`. When run, it actively monitors subnets and submits checkpoints.
-struct CheckpointSubsystem {
+pub struct CheckpointSubsystem {
     /// The subsystem uses a `ReloadableConfig` to ensure that, at all, times, the subnets under
     /// management are those in the latest version of the config.
     config: ReloadableConfig,
@@ -41,8 +41,7 @@ struct CheckpointSubsystem {
 
 impl CheckpointSubsystem {
     /// Creates a new `CheckpointSubsystem` with a configuration `config`.
-    #[allow(dead_code)]
-    fn new(config: ReloadableConfig) -> Self {
+    pub fn new(config: ReloadableConfig) -> Self {
         Self { config }
     }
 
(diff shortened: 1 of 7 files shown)


hmoniz:
Current dependencies on/for this PR:
* main
  * **PR #114** <a href="https://app.graphite.dev/github/pr/consensus-shipyard/ipc-agent/114" target="_blank">Graphite</a>  👈
    * **PR #115** <a href="https://app.graphite.dev/github/pr/consensus-shipyard/ipc-agent/115" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/consensus-shipyard/ipc-agent/114?utm_source=stack-comment).

src/server/handlers/config.rs
```diff
@@ -36,9 +32,9 @@ impl JsonRPCRequestHandler for ReloadConfigHandler {
     async fn handle(&self, request: Self::Request) -> anyhow::Result<Self::Response> {
         log::info!("received request to reload config: {request:?}");
 
-        let path = request
-            .path
-            .unwrap_or_else(|| self.default_config_path.clone());
-        self.config.reload(path).await
+        if request.path.is_some() {
+            self.config.set_path(request.path.unwrap());
```
adlrocha, line 36: Just to double-check, when we start the daemon if no config path is specified the default one is used to construct the `ReloadConfig` right?

adlrocha (approved):
LGTM, but I would like @cryptoAtwill to have a look before merging that he was the writing this piece of code. 

BTW, remember making the linter happy before merge :pray:

src/manager/checkpoint.rs
```diff
@@ -33,16 +33,15 @@ use crate::lotus::LotusClient;
...
     /// The subsystem uses a `ReloadableConfig` to ensure that, at all, times, the subnets under
     /// management are those in the latest version of the config.
     config: ReloadableConfig,
 }
 
 impl CheckpointSubsystem {
     /// Creates a new `CheckpointSubsystem` with a configuration `config`.
-    #[allow(dead_code)]
-    fn new(config: ReloadableConfig) -> Self {
+    pub fn new(config: ReloadableConfig) -> Self {
```
cryptoAtwill, line 44: Here should have `Arc` for the `ReloadableConfig`?

cryptoAtwill (approved):

src/manager/checkpoint.rs
hmoniz: Yes!

src/server/handlers/config.rs
hmoniz: It uses the one already present in `ReloadableConfig`.

PR #125, opened 2023-03-20
Update list subnet

cryptoAtwill:
Fix list subnets parsing error.

### DIFF
--- a/src/lotus/message/ipc.rs
+++ b/src/lotus/message/ipc.rs
@@ -41,7 +41,7 @@ pub struct IPCReadSubnetActorStateResponse {
 /// SubnetInfo is an auxiliary struct that collects relevant information about the state of a subnet
 ///
 /// Note that the serialization and deserialization casing are different. Reason because for deserialization,
-/// it is from the fvm actor, which is `PascalCase`. When serializa, we are using rust's default casing
+/// it is from the fvm actor, which is `PascalCase`. When serialize, we are using rust's default casing
 #[derive(Debug, Serialize, Deserialize)]
 pub struct SubnetInfo {
     /// Id of the subnet.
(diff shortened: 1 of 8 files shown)


src/lotus/message/ipc.rs
```diff
@@ -36,21 +39,28 @@ pub struct IPCReadSubnetActorStateResponse {
...
     /// Collateral staked in the subnet.
+    #[serde(rename(deserialize = "Stake"))]
     #[serde(deserialize_with = "deserialize_token_amount_from_str")]
+    #[serde(serialize_with = "serialize_token_amount_to_atto")]
     pub stake: TokenAmount,
     /// Circulating supply available in the subnet.
+    #[serde(rename(deserialize = "CircSupply"))]
     #[serde(deserialize_with = "deserialize_token_amount_from_str")]
-    #[serde(rename = "CircSupply")]
+    #[serde(serialize_with = "serialize_token_amount_to_atto")]
```
adlrocha, line 60: Why do we need this additional deserialization step? Can't we deserialize directly into token amount (and change SubnetInfo if needed)

src/lotus/message/tests.rs
```diff
@@ -78,3 +83,41 @@ fn test_token_amount_from_str() {
...
+#[test]
+fn test_subnet_info_from_str() {
+    set_current_network(Network::Testnet);
+
+    let raw_str = r#"
+    {
+        "ID": {
+            "Parent": "/root",
+            "Actor": "t01002"
+        },
```
adlrocha, line 109: We should check a grater value to see if it works with overflows (I had to fix a similar bug in deserialization  a few days ago

adlrocha (approved):
Nice. Thank you for the fix. Check the comments, see if they make sense and feel free to merge

src/lotus/message/ipc.rs
cryptoAtwill: TokenAmount does not implement `Serialize`, without the `serialize_with`, it's serialized into byte array which is definitely hard to read.

src/lotus/message/tests.rs
cryptoAtwill: yep, tested, it should be ok

PR #128, opened 2023-03-21
Several checkpoint-related fixes

adlrocha:
This PR introduces several fixes to checkpoints:
- Serialization of previous checkpoint from Lotus request.
- Serialization of `Checkpoint` from request to Lotus. Unfortunately, we can't use directly `gateway::Checkpoint` struct from actors for deserialization because it expects a tuple deserialization, while the serialization used by the Lotus RPC is a map serialization. 
- Improvements to the checkpoint logic. Mainly: 
  - Compute the right epoch to submit the checkpoint instead of waiting for the epoch to happen (in case we poll the blockchain when the checkpoint period has already passed). 
  - Check if the checkpoint for the epoch to be submitted has already been committed and do nothing if this is the case (preventing the process from failing, as re-committing a checkpoint returns an error in the actor). 
  - Wait synchronously for the checkpoint to be committed before moving on to the next iteration of the loop

As follow-up of this PR we should introduces additional fixes regarding the serialization of checkpoints. Mainly, we should have a `Checkpoint::from(CheckpointReponse)` that translates an Lotus RPC API checkpoint response into an actor serializable checkpoint object

### DIFF
--- a/src/manager/checkpoint.rs
+++ b/src/manager/checkpoint.rs
@@ -257,7 +257,7 @@ async fn manage_subnet((child, parent): (Subnet, Subnet), stop_notify: Arc<Notif
 
             // Sleep for an appropriate amount of time before checking the chain head again or return
             // if a stop notification is received.
-            if wait_next_iteration(&stop_notify).await? {
+            if !wait_next_iteration(&stop_notify).await? {
                 return Ok(());
             }
         }


src/manager/checkpoint.rs
```diff
@@ -243,6 +268,15 @@ async fn manage_subnet((child, parent): (Subnet, Subnet), stop_notify: Arc<Notif
     ))
 }
 
+/// Sleeps for some time if stop_notify is not fired. It returns true to flag that we should move to the
+/// next iteration of the loop, while false informs that the loop should return.
+async fn wait_next_iteration(stop_notify: &Arc<Notify>) -> Result<bool> {
+    select! {
+        _ = sleep(CHAIN_HEAD_REQUEST_PERIOD) => {Ok(true)}
+        _ = stop_notify.notified() => {Ok(false)}
```
cryptoAtwill, line 276: not super familiar with the `notified` api, I could be wrong, but just curious will there be missed signal. Consider the reloadable config had an update in config, then causing the `stop_notify` to invoke `notify_all`, but the checkpointing thread has not yet reached `wait_next_iteration`, and when reaching select, what happens to `stop_notify.notified()`, will it receive the signal or miss?

cryptoAtwill (approved):

src/manager/checkpoint.rs
```diff
@@ -170,70 +177,88 @@ async fn manage_subnet((child, parent): (Subnet, Subnet), stop_notify: Arc<Notif
...
                     }
                 }
+            };
+
+            // Now, for each account defined in the `child` subnet that is in the validator set, we
+            // submit a checkpoint on its behalf.
+            assert_eq!(child_head.cids.len(), 1); // Again, check key assumption
+            let child_tip_set = Cid::try_from(child_head.cids.first().unwrap().clone())?;
+            for account in child.accounts.iter() {
+                if validator_set.contains(account) {
```
cryptoAtwill, line 238: what happens if none of the accounts is in the `validator_set`, we are empty spining? or we assume there will be an account in the validator set?

src/manager/checkpoint.rs
```diff
@@ -170,70 +177,88 @@ async fn manage_subnet((child, parent): (Subnet, Subnet), stop_notify: Arc<Notif
...
+                    .await?;
+                }
             }
 
             // Sleep for an appropriate amount of time before checking the chain head again or return
             // if a stop notification is received.
-            select! {
-                _ = sleep(CHAIN_HEAD_REQUEST_PERIOD) => {}
-                _ = stop_notify.notified() => { return Ok(()); }
+            if wait_next_iteration(&stop_notify).await? {
```
cryptoAtwill, line 260: I think here should be `!wait_next_iteration(&stop_notify)`?

adlrocha: Good call, completely missed it. Thanks!

src/manager/checkpoint.rs
adlrocha: You just do nothing for this round of checkpointing. You are just waiting in the background for when is your turn to submit a new checkpoint

src/manager/checkpoint.rs
adlrocha: @hmoniz, can you comment on this? TBH, I didn't write this logic and I would have to look deeper into it

PR #129, opened 2023-03-22
Prototype of topdown message submission

hmoniz:
This is just for getting early feedback. Not for merging.

### DIFF
--- a/src/manager/crossmsgs.rs
+++ b/src/manager/crossmsgs.rs
@@ -1,0 +1,50 @@
+// Copyright 2022-2023 Protocol Labs
+// SPDX-License-Identifier: MIT
+use anyhow::Result;
+use fvm_shared::address::Address;
+use fvm_shared::clock::ChainEpoch;
+use ipc_gateway::CrossMsg;
+use crate::jsonrpc::JsonRpcClient;
+
+use crate::lotus::client::LotusJsonRPCClient;
+
+// Prototype function for submitting topdown messages. This function is supposed to be called each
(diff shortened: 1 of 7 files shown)


src/manager/crossmsgs.rs
```diff
@@ -0,0 +1,58 @@
...
+use crate::lotus::LotusClient;
+
+// Prototype function for submitting topdown messages. This function is supposed to be called each
+// Nth epoch of a parent subnet. It reads the topdown messages from the parent subnet and submits
+// them to the child subnet.
+async fn submit_topdown_msgs<T: JsonRpcClient + Send + Sync>(
+    parent_epoch: ChainEpoch,
+    account: &Address,
+    child_subnet: SubnetID,
+    child_client: &LotusJsonRPCClient<T>,
```
aakoshh, line 27: It would be great to have an abstraction between the manager and Lotus, to work with your own domain types, and not assume that every child subnet implements Lotus' API.

src/manager/crossmsgs.rs
```diff
@@ -0,0 +1,58 @@
...
+    parent_client: &LotusJsonRPCClient<T>,
```
aakoshh, line 28: Same goes for the parent, to support recursive subnets with anything other than Lotus.

adlrocha:
@hmoniz, I think this one no longer applies. Feel free to re-open it if this is the case (I won't delete the branch until you confirm, feel free to delete it yourself).

PR #131, opened 2023-03-24
Adds a list-checkpoints command

hmoniz:
Adds a command that lists a range of checkpoints. This is predicated on a Lotus JSON-RPC method for getting a range of checkpoints.

### DIFF
--- a/src/cli/commands/manager/list_checkpoints.rs
+++ b/src/cli/commands/manager/list_checkpoints.rs
@@ -8,6 +8,7 @@ use fvm_shared::clock::ChainEpoch;
 use ipc_sdk::subnet_id::SubnetID;
 use std::collections::HashMap;
 use std::fmt::Debug;
+use ipc_gateway::Checkpoint;
 
 use crate::cli::commands::get_ipc_agent_url;
 use crate::cli::{CommandLineHandler, GlobalArguments};
@@ -36,7 +37,7 @@ impl CommandLineHandler for ListCheckpoints {
         };
 
         let checkpoints = json_rpc_client
-            .request::<HashMap<SubnetID, SubnetInfo>>(
+            .request::<HashMap<SubnetID, Checkpoint>>(
                 json_rpc_methods::LIST_CHECKPOINTS,
                 serde_json::to_value(params)?,
             )
--- a/src/cli/commands/manager/list_checkpoints.rs
+++ b/src/cli/commands/manager/list_checkpoints.rs
@@ -2,16 +2,17 @@
 // SPDX-License-Identifier: MIT
 //! List checkpoints cli command
 
+use std::fmt::Debug;
+
 use async_trait::async_trait;
 use clap::Args;
 use fvm_shared::clock::ChainEpoch;
-use ipc_gateway::Checkpoint;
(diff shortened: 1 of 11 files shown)


src/cli/commands/manager/list_checkpoints.rs
```diff
@@ -0,0 +1,62 @@
...
+        let json_rpc_client = JsonRpcClientImpl::new(url, None);
+
+        let params = ListCheckpointsParams {
+            subnet_id: arguments.subnet_id.clone(),
+            from_epoch: arguments.from_epoch,
+            to_epoch: arguments.to_epoch,
+        };
+
+        let checkpoints = json_rpc_client
+            .request::<HashMap<SubnetID, SubnetInfo>>(
```
cryptoAtwill, line 39: This `SubnetInfo` should be `Checkpoint`?

cryptoAtwill (changes requested):

src/cli/commands/manager/list_checkpoints.rs
hmoniz: Yes, fixed

src/cli/commands/manager/list_checkpoints.rs
```diff
@@ -0,0 +1,62 @@
...
+        let json_rpc_client = JsonRpcClientImpl::new(url, None);
+
+        let params = ListCheckpointsParams {
+            subnet_id: arguments.subnet_id.clone(),
+            from_epoch: arguments.from_epoch,
+            to_epoch: arguments.to_epoch,
+        };
+
+        let checkpoints = json_rpc_client
+            .request::<HashMap<SubnetID, Checkpoint>>(
```
adlrocha, line 39: Typo? The key of the hashmap should be the epoch, right? 
```suggestion
            .request::<HashMap<Epoch, Checkpoint>>(
```

adlrocha (changes requested):
It was a bit hard to review, I think it picked up many changes from the base branch that you used for some reason (which have already merged to main). 

Overall, LGTM, but for it to work you'll need to incorporate the changes from this PR: https://github.com/consensus-shipyard/ipc-agent/pull/136. The current way we are using to deserialize the list checkpoints wouldn't work, unfortunately.

src/lotus/client.rs
```diff
@@ -337,6 +399,32 @@ impl<T: JsonRpcClient + Send + Sync> LotusClient for LotusJsonRPCClient<T> {
...
+            {
+                "Parent": parent,
+                "Actor": actor
+            },
+            from_epoch,
+            to_epoch
+        ]);
+        let r = self
+            .client
+            .request::<Vec<Checkpoint>>(methods::IPC_LIST_CHECKPOINTS, params)
```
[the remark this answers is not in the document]
hmoniz: Fixed

src/cli/commands/manager/list_checkpoints.rs
hmoniz: Fixed. This actually should be Vec<Checkpoint>

adlrocha (approved):
Tested and fixed sync

PR #137, opened 2023-03-28
Include gateway addr in subnet config

cryptoAtwill:
The changes in this PR:
- Add gateway address and network name in subnet config. 
- Replace subnets from `Hashmap<String, Subnet>` to `Hashmap<SubnetID, Subnet>`, so that all the look ups we can directly use subnet id instead of converting to string.
- Updated subnets config from key value map style to array.

### DIFF
--- a/config/template.toml
+++ b/config/template.toml
@@ -3,14 +3,18 @@ json_rpc_address = "127.0.0.1:3030"
 
 [subnets]
 
-[subnets.root]
+[[subnets]]
 id = "/root"
+network_name = "root"
+gateway_addr = 64
 jsonrpc_api_http = "https://example.org/rpc/v0"
 jsonrpc_api_ws = "wss://example.org/rpc/v0"
 auth_token = "YOUR ROOT AUTH TOKEN"
 
-[subnets.child]
+[[subnets]]
 id = "/root/t0100"
+network_name = "child"
+gateway_addr = 64
 jsonrpc_api_http = "https://example.org/rpc/v2"
 auth_token = "YOUR CHILD AUTH TOKEN"
 accounts = ["t3thgjtvoi65yzdcoifgqh6utjbaod3ukidxrx34heu34d6avx6z7r5766t5jqt42a44ehzcnw3u5ehz47n42a"]
--- a/src/manager/checkpoint.rs
+++ b/src/manager/checkpoint.rs
@@ -123,13 +123,7 @@ impl IntoSubsystem<anyhow::Error> for CheckpointSubsystem {
 /// manage checkpoint for. This means that for each `child_subnet` there exists at least one account
 /// for which we need to submit checkpoints on behalf of to `parent_subnet`, which must also be
 /// present in the map.
-fn subnets_to_manage(subnets: &HashMap<String, Subnet>) -> Vec<(Subnet, Subnet)> {
-    // First, we remap subnets by SubnetID.
-    let subnets_by_id: HashMap<SubnetID, Subnet> = subnets
-        .values()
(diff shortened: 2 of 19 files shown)


src/manager/checkpoint.rs
```diff
@@ -123,13 +123,7 @@ impl IntoSubsystem<anyhow::Error> for CheckpointSubsystem {
...
 /// present in the map.
-fn subnets_to_manage(subnets: &HashMap<String, Subnet>) -> Vec<(Subnet, Subnet)> {
-    // First, we remap subnets by SubnetID.
-    let subnets_by_id: HashMap<SubnetID, Subnet> = subnets
-        .values()
-        .map(|s| (s.id.clone(), s.clone()))
-        .collect();
-
+fn subnets_to_manage(subnets_by_id: &HashMap<SubnetID, Subnet>) -> Vec<(Subnet, Subnet)> {
     // Then, we filter for subnets that have at least one account and for which the parent subnet
```
hmoniz, line 127: nit: Update this comment to remove the "Then"

hmoniz (approved):

src/manager/checkpoint.rs
cryptoAtwill: Removed and resolving this.

config/template.toml
```diff
@@ -3,14 +3,18 @@ json_rpc_address = "127.0.0.1:3030"
...
+[[subnets]]
 id = "/root"
+network_name = "root"
+gateway_addr = 64
 jsonrpc_api_http = "https://example.org/rpc/v0"
 jsonrpc_api_ws = "wss://example.org/rpc/v0"
 auth_token = "YOUR ROOT AUTH TOKEN"
 
-[subnets.child]
+[[subnets]]
```
adlrocha, line 14: From total ignorance, would it be possible to make the config something like so that we use the `id` already as the key of the map?: 
```
[subnets]

[[/root]]
....
[[/root/t01011]]
```

adlrocha (changes requested):
A few minor comments (they are non-blocking, so feel free to disregard them if you don't feel they apply right now). 

And the reason for the request for changes. IIRC we were using `DEFAULT_GATEWAY_ADDR` all over the place whenver there was a request that required the gateway as an argument. I don´t see that changed in the code, did I miss it or should we change that also before we merge and pick up the gateway address from the config.

config/template.toml
```diff
@@ -3,14 +3,18 @@ json_rpc_address = "127.0.0.1:3030"
 
 [subnets]
 
-[subnets.root]
+[[subnets]]
 id = "/root"
+network_name = "root"
+gateway_addr = 64
```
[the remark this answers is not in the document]
cryptoAtwill: perhaps, it's just convenience I guess. Convert actor id to address is quite straightforward. If we want to use string, it works too.

config/template.toml
cryptoAtwill: This one I'm not so sure, but I think it cannot be done, because `[[/root]]` would translate to hashmap of "/root" as key and whatever in the `...` as an array of deserialized struct. But I think we should avoid using subnet id string because of `/`, everytime we have to escape it with `"`, which is easy to make mistake.

config/template.toml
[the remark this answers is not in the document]
adlrocha: Let's either use `gateway_addr = "t064"` or let's call this field `gateway_id` and make it an uint. Whatever works best for you.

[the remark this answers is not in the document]
cryptoAtwill: use the first option, please check

adlrocha (approved):

PR #144, opened 2023-03-31
fix config multi address error

cryptoAtwill:
Fix config multi address error. `visit_seq` does not seems to work.


adlrocha (approved):
LGTM. @hmoniz, you wrote this bit, any concern with @cryptoAtwill's approach removing the `SeqVisitor`? Thanks!

hmoniz:
> LGTM. @hmoniz, you wrote this bit, any concern with @cryptoAtwill's approach removing the `SeqVisitor`? Thanks!

If it works, I'm good with it!

PR #156, opened 2023-04-12
Adapt bottom-up checkpoints to actor refactor

adlrocha:
This PR adapts the logic of bottom-up checkpoint submission to the new actor refactor.

### DIFF
--- a/src/lotus/client.rs
+++ b/src/lotus/client.rs
@@ -22,9 +22,7 @@ use serde_json::json;
 use crate::jsonrpc::{JsonRpcClient, JsonRpcClientImpl, NO_PARAMS};
 use crate::lotus::json::ToJson;
 use crate::lotus::message::chain::ChainHeadResponse;
-use crate::lotus::message::ipc::{
-    IPCReadGatewayStateResponse, IPCReadSubnetActorStateResponse, Votes,
-};
+use crate::lotus::message::ipc::{IPCReadGatewayStateResponse, IPCReadSubnetActorStateResponse};
 use crate::lotus::message::mpool::{
     MpoolPushMessage, MpoolPushMessageResponse, MpoolPushMessageResponseInner,
 };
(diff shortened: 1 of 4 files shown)


src/lotus/client.rs
```diff
@@ -47,13 +45,13 @@ mod methods {
...
-    pub const IPC_GET_CHECKPOINT_TEMPLATE: &str = "Filecoin.IPCGetCheckpointTemplate";
-    pub const IPC_GET_CHECKPOINT: &str = "Filecoin.IPCGetCheckpoint";
+    pub const IPC_GET_CHECKPOINT_TEMPLATE: &str = "Filecoin.IPCGetCheckpointTemplateSerialized";
+    pub const IPC_GET_CHECKPOINT: &str = "Filecoin.IPCGetCheckpointSerialized";
     pub const IPC_READ_GATEWAY_STATE: &str = "Filecoin.IPCReadGatewayState";
     pub const IPC_READ_SUBNET_ACTOR_STATE: &str = "Filecoin.IPCReadSubnetActorState";
     pub const IPC_LIST_CHILD_SUBNETS: &str = "Filecoin.IPCListChildSubnets";
-    pub const IPC_GET_VOTES_FOR_CHECKPOINT: &str = "Filecoin.IPCGetVotesForCheckpoint";
-    pub const IPC_LIST_CHECKPOINTS: &str = "Filecoin.IPCListCheckpoints";
+    pub const IPC_VALIDATOR_HAS_VOTED_BOTTOMUP: &str = "Filecoin.IPCHasVotedBottomUpCheckpoint";
```
hmoniz, line 53: Can we eventually add one for TopDown as well?

src/manager/checkpoint.rs
```diff
@@ -164,87 +164,103 @@ async fn manage_subnet((child, parent): (Subnet, Subnet), stop_notify: Arc<Notif
...
+                            .ipc_validator_has_voted_bottomup(&child.id, epoch, account)
+                            .await
+                            .map_err(|e| {
+                                log::error!(
+                                    "error checking if validator has voted in subnet: {:?}",
+                                    &child.id
+                                );
+                                e
+                            })?;
+                        if !has_voted {
```
hmoniz, line 219: Instead of this, maybe we can take the `child.accounts` vector and filter it first in a functional way for the accounts that are in the validator set and haven't voted.

src/manager/checkpoint.rs
```diff
@@ -164,87 +164,103 @@ async fn manage_subnet((child, parent): (Subnet, Subnet), stop_notify: Arc<Notif
...
+                                child_tip_set,
+                                submission_epoch,
+                                account,
+                                &child,
+                                &child_client,
+                                &parent_client,
+                            )
+                            .await?;
+
+                            // check if by any chance we have the opportunity to submit any outstanding checkpoint we may be
```
hmoniz, line 231: Is this really necessary? Can't we get away by simply moving forward and submitting a checkpoint for the latest epoch? I'm assuming that by checkpointing some epoch e, we're also implicitly checkpointing all previous epochs.

hmoniz (approved):

src/lotus/client.rs
adlrocha: Oh, yes. This is already implemented and tested in the Lotus side. I will add it to this PR, so we can then merge with yours.

src/manager/checkpoint.rs
adlrocha: This one is tricky. While theoretically I agree this is the case, practically we don't implement a way to implicitly checkpoint all the "previous history" in the actor because each checkpoint is populated with its own list of bottom-up messages, and we don't know which ones have been executed already in the parent. 

And while I feel that with top-down messages this would be possible, we currently also implement the voting sequentially, where all the checkpoints are expected to be submitted sequentially. Actually, we should double-check that this is the way we are submitting them in top-down checkpoints.

src/manager/checkpoint.rs
adlrocha: Let's discuss this one at some point, I may not be proficient enough in Rust, but I can't seem to be able to make  async blocks work inside `filter`. Once I have a bit more time to breath I will dig a bit more

hmoniz: No worries. It was just a nitpick suggestion.

Issue #91, opened 2023-03-10
Tests: End-to-end integration tests


adlrocha:
Implement a test framework for the IPC agent to perform end-to-end tests of IPC. 

See [design doc ](https://docs.google.com/document/d/14lkRRv6MQYnuEfp2GoGngdD8Q5YgfE38D8HTZWKgKf4/edit#heading=h.ow3dln7nemch)for further details.

adlrocha:
This PR includes a first skeleton of how to approach integration tests: https://github.com/consensus-shipyard/ipc-agent/pull/143

adlrocha:
A good first foundation for this has been implemented in #157, 161 and https://github.com/consensus-shipyard/ipc-agent/pull/206

PR #157, opened 2023-04-13
IPC-91: Test env

aakoshh:
Part of #91 

Sets up an environment with `docker-compose` running agents and nodes.


testing/e2e/README.md
```diff
@@ -9,3 +11,46 @@ cargo test -p ipc_e2e --test subnet_lifecycle
...
+
+5 directories, 7 files
+```
+
+There can be multiple agents, and their corresponding `config.toml` files will be generated as we create more nodes and subnets. To create another agent, we would run `make agent IPC_AGENT_NR=1`.
+
+The main targets of the `Makefile` are:
+
+* `make agent`: create a configuration directory for `$IPC_AGENT_NR`; the container isn't started yet, so we could make some modifications if necessary
+* `make agent/up`: start the docker container for `$IPC_AGENT_NR`; if necessary build the `ipc-agent` docker image, the configuration directory, etc.
```
adlrocha, line 52: Quick sanity-check: does the script check the commit it is on and re-generates the image automatically if the commit changes, or this should be manually by users? I am just wondering if we should add a note here around that.

adlrocha (approved):
LGTM. This is going to be really really useful for testing. Thanks!

testing/e2e/README.md
aakoshh: At he moment the dependency is here:
https://github.com/consensus-shipyard/ipc-agent/blob/1749ce37556631ee90498a353fcc1007797d71b9/testing/e2e/Makefile#L177

which depends on the agent image: 
https://github.com/consensus-shipyard/ipc-agent/blob/1749ce37556631ee90498a353fcc1007797d71b9/testing/e2e/Makefile#L270

Which depends on all the Rust source. 

Because the dependency is this form:
```
$(IPC_AGENT_DIR)/steps/up: .make/docker/network .make/docker/images/ipc-agent
```
and not this with the extra `|`:
```
$(IPC_AGENT_DIR)/steps/up: | .make/docker/network .make/docker/images/ipc-agent
```
I expect that it would indeed re-build the image and redeploy the agent.

Whether that's expected and valuable, I don't know. I think with the agent there's no harm in it, and if the same thing happened to the node, hopefully the mounted `repo_directory` volume would survive.

PR #158, opened 2023-04-14
Update guide and docs

jsoares:
* The guide is supposed to go on the docs website, so it's here for review and awaiting docs merging
  * It's supposed to be a really quick path towards having something working, so it skips on the explanations and can't handle exceptions. But the goal is to have something that works reliably on a clean setup and allows for testing.
* The readme has been stripped to just build the agent and get it working with a rootnet, with everything else moving to subdocs
  * Also made an effort to simplify things across the board
  * These subdocs can stay here or move to the docs website in some form
* I tried to harmonise a number of things, e.g. code blocks, argument passing with = or without, etc. but probably didn't go all the way.
* I think I also replaced all ipc-agent commands with the new grouped variants but may have missed something
* **The guide assumes that the default branches will work. That's not currently true so someone following the guide will fail. It will hopefully be true on Monday.**


TippyFlitsUK (approved):

jsoares:
One note: The subnet deployment instructions (under docs/subnet) are very duplicative for the single- and multi-node case... I think we should consider dropping the single node instructions; other than some minor simplifying aspects (e.g. sharing default wallet), it's just a special case of multinode.

adlrocha (approved):
This is great! Thank you @jsoares

PR #160, opened 2023-04-14
Top-down Checkpoints

hmoniz:
Implementation of top-down checkpointing. This also refactors (slightly) the checkpoint manager to be able to handle both bottom-up and top-down checkpoints. There's a bit of duplication between managing bottom-up and top-down checkpoints, but we'll leave that refactoring after we've tested this implementation.


cryptoAtwill (changes requested):

src/manager/topdown.rs
```diff
@@ -0,0 +1,247 @@
...
+    // FIXME: We probably shouldn't use the default one here and use the one
+    // deployed in the subnet
+    let gateway_addr = Address::from_str(GATEWAY_ACTOR_ADDRESS)?;
+    let submission_tip_set = parent_client
+        .get_tipset_by_height(submission_epoch, curr_parent_tip_set)
+        .await?;
+    let cid_map = submission_tip_set.cids.first().unwrap().clone();
+    let submission_tip_set = Cid::try_from(cid_map)?;
+    let top_down_msgs = parent_client
+        .ipc_get_topdown_msgs(&child_subnet, gateway_addr, submission_tip_set, nonce)
```
cryptoAtwill, line 211: do we need some kind of pagination support to limit the return size? If there are a lot of top down msgs, it would blow up the gas?

src/manager/topdown.rs
```diff
@@ -0,0 +1,247 @@
...
+        .await?;
```
cryptoAtwill, line 212: btw, what's the relationship between `nonce`, `submission_tip_set`, `curr_parent_tip_set` and `curr_child_tip_set`. Do they need to have some relationship to get the correct messages? Or just latest tipset or nonce would do?

cryptoAtwill (changes requested):

src/manager/topdown.rs
```diff
@@ -0,0 +1,247 @@
...
+                            let cid_map = child_head.cids.first().unwrap().clone();
+                            let child_tip_set = Cid::try_from(cid_map)?;
+                            let child_gw_state =
+                                child_client.ipc_read_gateway_state(child_tip_set).await?;
+                            let last_exec = child_gw_state
+                                .top_down_checkpoint_voting
+                                .last_voting_executed;
+                            let submission_epoch = last_exec + period;
+                            if curr_epoch >= submission_epoch {
+                                submit_topdown_checkpoint(
```
cryptoAtwill, line 147: If the validator is lagging the current epoch, we have to speed up the submission. Otherwise, the checkpoint would be processed too slow.

src/manager/topdown.rs
```diff
@@ -0,0 +1,155 @@
...
+    // submit to the child subnet.
+    let child_head = parent_client.chain_head().await?;
+    let cid_map = child_head.cids.first().unwrap().clone();
+    let child_tip_set = Cid::try_from(cid_map)?;
+    let state = child_client.ipc_read_gateway_state(child_tip_set).await?;
+    let nonce = state.applied_topdown_nonce + 1;
+
+    // Then, we read from the parent subnet the topdown messages with nonce greater than or equal
+    // to the nonce we just obtained.
+    let gateway_addr = Address::from_str(GATEWAY_ACTOR_ADDRESS)?;
```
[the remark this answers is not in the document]
adlrocha: Yes. I found other places in the code that are already using this one, so I will add an issue to fix throughout the whole code. Resolving for now.

src/manager/topdown.rs
adlrocha: That's a great point. I will also track this one in an issue. I think this is a problem for the future, but we will definitely have to use some kind of logic to prevent from running out of gas inside a block. The thing is that we need this to be deterministic so all checkpoints include the same messages, if not this can end up in checkpoints never being executed.

src/manager/topdown.rs
adlrocha: Discussed sync. TL;DR, we get the current tipsets for parent and child, and then use the submission epoch at the child to populate the checkpoint.

src/manager/topdown.rs
adlrocha: Already fixed with the loop. This will be improved further with the refactor.

PR #161, opened 2023-04-14
IPC-91: Test environment (Part 2)

aakoshh:
Part of #91 

Continuing down the README from where https://github.com/consensus-shipyard/ipc-agent/pull/157 left off. 

New commands in this PR:
* `make topologies/<name>.sh` to turn `topologies/<name>.yaml` into a runnable script
* `make setup/<name>` to compile and run a topology script
* `make wallet` creates a new wallet
* `make subnet` creates a new subnet
* `make wallet/fund` sends tokens to the wallet
* `make subnet/fund` sends tokens to the subnet, executed on the parent node
* `make subnet/join` makes the validator join the subnet , executed on the parent node

The only sane way to run these is e.g. `make setup/simple`. 

### Notes 

It's not expected to start a validator before joining a subnet, yet it's done to simplify starting the containers with docker-compose. The validator will crash the first time it's started, but it's given a kick again in the `post-up` step, after joining; at that point it runs.


aakoshh:
Useful tip from Will:

```❯ docker exec -it ipc-node-0-daemon eudico wallet list
Address                                    Balance                      Nonce  Default  
t1ah7phdlf5eg2nf25trxvvddzzde7w6ehtrfz2yi  1 FIL                        0               
t1cp4q4lqsdhob23ysywffg2tvbmar5cshia4rweq  4998.998985909661230076 FIL  2      X    
```

aakoshh:
Here's an example log from setting up the simple topology: [setup-log.txt](https://github.com/consensus-shipyard/ipc-agent/files/11330676/setup-log.txt)

It actually fails at the end while setting up the sub-subnet with `actor not found`. I don't know why this error happens, but it's very frequent with the sub-subnet. cc @adlrocha

adlrocha:
> It's not expected to start a validator before joining a subnet, yet it's done to simplify starting the containers with docker-compose. The validator will crash the first time it's started, but it's given a kick again in the post-up step, after joining; at that point it runs.

We are improving this after this PR, and instead of crashing the validator polls to see if it can start mining on a subnet or not. https://github.com/consensus-shipyard/lotus/issues/166

src/cli/commands/mod.rs
```diff
@@ -96,13 +96,7 @@ pub async fn cli() {
...
     };
 
-    if let Err(e) = r {
-        log::error!(
-            "process command: {:?} failed due to error: {:?}",
-            args.command,
-            e
-        )
-    }
+    r.with_context(|| format!("error processing command {:?}", args.command))
```
adlrocha, line 99: This is really helpful, thanks!

adlrocha (commented):
:ship: This a perfect first iteration of what we wanted for the integration test. I think it can really reduce the debugging time. Let's get this merged :)

src/cli/commands/mod.rs
aakoshh: Like I said the other change I would make (one that would probably break the scripts) is to [log to stderr](https://github.com/consensus-shipyard/fendermint/blob/master/fendermint/app/src/main.rs#L22) (which the agent does at the moment already) and [use simple println](https://github.com/consensus-shipyard/fendermint/blob/master/fendermint/app/src/cmd/rpc.rs#L221) to output results. That would make it easier to automate.

PR #167, opened 2023-04-18
docs for cross-net messages

adlrocha:
This PR includes some docs for the usage and operation of cross-net messages.

### DIFF
--- a/docs/subnet.md
+++ b/docs/subnet.md
@@ -25,7 +25,7 @@ If your daemon is running on a docker container, you can get the container id or
 $ docker exec -it <container-id> eudico wallet export --lotus-json <adress-to-export> > ~/.ipc-agent/wallet.key
 
 # Example execution
-$ docker exec -it 84711d67cf162e30747c4525d69728c4dea8c6b4b35cd89f6d0947fee14bf908 eudico wallet export --lotus-json t1cp4q4lqsdhob23ysywffg2tvbmar5cshia4rweq > ~/.ipc-agent/wallet.key
+$ docker exec -it ipc_root_1234 eudico wallet export --lotus-json t1cp4q4lqsdhob23ysywffg2tvbmar5cshia4rweq > ~/.ipc-agent/wallet.key
 ```
 
 ### Importing wallet keys
@@ -37,7 +37,7 @@ Depending on whether the subnet is running inside a docker container or not, you
 $ ./eudico wallet import --lotus-json <wallet-key-file-path>
 
 # Example execution
-$ ./eudico wallet import --lotus-json ~/.ipc-agent/t1ivy6mo2ofxw4fdmft22nel66w63fb7cuyslm4cy.key
+$ ./eudico wallet import --lotus-json ~/.ipc-agent/wallet.key
 ```
 
 ```bash
@@ -45,7 +45,7 @@ $ ./eudico wallet import --lotus-json ~/.ipc-agent/t1ivy6mo2ofxw4fdmft22nel66w63
 $ docker cp <wallet-key-path> <container-id>:<target-file-in-container> && docker exec -it <container-id> eudico wallet import --format=json-lotus <target-file-in-container>
 
 # Example execution
-$ docker cp ~/.ipc-agent/t1ivy6mo2ofxw4fdmft22nel66w63fb7cuyslm4cy.key 91d2af805346:/input.key && docker exec -it 91d2af805346 eudico wallet import --format=json-lotus input.key
+$ docker cp ~/.ipc-agent/wallet.key ipc_root_t01002_1250:/input.key && docker exec -it ipc_root_t01002_1250 eudico wallet import --format=json-lotus input.key
 ```
 
 ## Running a simple subnet with a single validator
@@ -57,13 +57,13 @@ This section provides instructions for spawning a simple subnet with a single va
 To run a subnet the first thing is to configure and create the subnet actor that will govern the subnet's operation.
 
 ```bash
-$ ./bin/ipc-agent subnet create -p <parent-id> -n <subnet-name> --min-validator-stake 1 --min-validators <num-validators> --finality-threshold <number-epochs> --check-period <epochs-between-checks>
+$ ./bin/ipc-agent subnet create --parent <parent> --name <name> --min-validator-stake <min_validator_stake> --min-validators <min-validators> --bottomup-check-period <bottomup-check-period> --topdown-check-period <topdown-check-period>
 
 # Example execution
-$ ./bin/ipc-agent subnet create -p /root -n test --min-validator-stake 1 --min-validators 0 --finality-threshold 10 --check-period 10
+$ ./bin/ipc-agent subnet create --parent /root --name test --min-validator-stake 1 --min-validators 0 --bottomup-check-period 30 --topdown-check-period 30
 [2023-03-21T09:32:58Z INFO  ipc_agent::cli::commands::manager::create] created subnet actor with id: /root/t01002
 ```
-This command deploys a subnet actor for a new subnet from the `root`, with a human-readable name `test`, that requires at least `1` validator to join the subnet to be able to mine new blocks, and with a checkpointing period to the parent of `10` blocks. We can see that the output of this command is the ID of the new subnet.
+This command deploys a subnet actor for a new subnet from the `root`, with a human-readable name `test`, that requires at least `1` validator to join the subnet to be able to mine new blocks, and with a checkpointing period (both bottom-up and top-down) of `30` blocks. We can see that the output of this command is the ID of the new subnet.
 
 ### Exporting your wallet
 
@@ -139,7 +139,7 @@ $ ./bin/ipc-agent subnet send-value --subnet /root --to <target-wallet> <amount_
 
 With this, we can already create the subnet with `/root` as its parent. We are going to set the `--min-validators 2` so no new blocks can be created without this number of validators in the subnet.
 ```bash
-./bin/ipc-agent subnet create -p /root -n test --min-validator-stake 1 --min-validators 2 --finality-threshold 10 --check-period 10
+./bin/ipc-agent subnet create --parent /root --name test --min-validator-stake 1 --min-validators 2 --bottomup-check-period 30 --topdown-check-period 30
 ```
 ### Deploying the infrastructure
 
--- a/docs/subnet.md
+++ b/docs/subnet.md
@@ -67,12 +67,7 @@ This command deploys a subnet actor for a new subnet from the `root`, with a hum
 
 ### Exporting your wallet
 
-Let's export the default wallet (or other wallet you'd like to use) for use in the subnet validator.
-```bash
-$ ./eudico wallet export --lotus-json `./eudico wallet default` > ~/.ipc-agent/wallet.key
-```
-
-Make sure that your wallet has enough funds to put up the collateral to join the subnet.
+We will need to export the wallet key from our root node so that we can import them to our validators. Depending on how you are running your rootnet node you'll have to make a call to the docker container, or your nodes API. More information about exporting keys from your node can be found under [this section](#Exporting-wallet-keys). Make sure that the wallet holds enough funds to meet the subnet collateral requirements.
 
 ### Deploying a subnet node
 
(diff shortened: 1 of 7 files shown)


docs/subnet.md
```diff
@@ -57,22 +57,17 @@ This section provides instructions for spawning a simple subnet with a single va
 To run a subnet the first thing is to configure and create the subnet actor that will govern the subnet's operation.
 
 ```bash
-$ ./bin/ipc-agent subnet create -p <parent-id> -n <subnet-name> --min-validator-stake 1 --min-validators <num-validators> --finality-threshold <number-epochs> --check-period <epochs-between-checks>
+$ ./bin/ipc-agent subnet create --parent <parent> --name <name> --min-validator-stake <min_validator_stake> --min-validators <min-validators> --bottomup-check-period <bottomup-check-period> --topdown-check-period <topdown-check-period>
```
jsoares, line 60: These commands were broken due to wrong arguments. I fixed them in this repo but was wondering whether they might be found elsewhere. I don't think so, but just a note.

docs/subnet.md
```diff
@@ -57,22 +57,17 @@ This section provides instructions for spawning a simple subnet with a single va
...
 
 # Example execution
-$ ./bin/ipc-agent subnet create -p /root -n test --min-validator-stake 1 --min-validators 0 --finality-threshold 10 --check-period 10
+$ ./bin/ipc-agent subnet create --parent /root --name test --min-validator-stake 1 --min-validators 0 --bottomup-check-period 30 --topdown-check-period 30
 [2023-03-21T09:32:58Z INFO  ipc_agent::cli::commands::manager::create] created subnet actor with id: /root/t01002
 ```
-This command deploys a subnet actor for a new subnet from the `root`, with a human-readable name `test`, that requires at least `1` validator to join the subnet to be able to mine new blocks, and with a checkpointing period to the parent of `10` blocks. We can see that the output of this command is the ID of the new subnet.
+This command deploys a subnet actor for a new subnet from the `root`, with a human-readable name `test`, that requires at least `1` validator to join the subnet to be able to mine new blocks, and with a checkpointing period (both bottom-up and top-down) of `30` blocks. We can see that the output of this command is the ID of the new subnet.
```
jsoares, line 66: It's worth noting that the unit of the checkpointing period isn't expressed in the --help; one could wonder if it's second or block periods.

docs/subnet.md
jsoares, line 60: These commands were broken due to wrong arguments. I fixed occurrences in this repo but was wondering whether they might be found elsewhere. I don't think so, but just a note.

docs/usage.md
```diff
@@ -27,6 +27,9 @@ $ ./bin/ipc-agent subnet join --subnet /root/t01002 --collateral 2 --validator-n
 ```
 This command specifies the subnet to join, the amount of collateral to provide and the validator net address used by other validators to dial them.
 
+## Listing your balance in a subnet 
+> TODO: Add a note on how `wallet list` can be used to check your funds in a subnet. This should be implemented tomorrow (if things go well).
```
jsoares, line 31: TODO

docs/usage.md
```diff
@@ -37,28 +40,75 @@ $ ./bin/ipc-agent subnet send-value --subnet <subnet-id> --to <to-addr> <value>
...
+# Example execution
+$ ./bin/ipc-agent cross-msg release --subnet=/root/t01002 100
+```
+This command includes the cross-net message into a bottom-up checkpoint after the current epoch. Once the bottom-up checkpoint is committed, you should see the funds in your account in the parent. 
 
-### Listing checkpoints from a subnet
+
+## Listing checkpoints from a subnet
 
 Subnets are periodically committing checkpoints to their parent every `check-period` (parameter defined when creating the subnet). If you want to inspect the information of a range of checkpoints committed in the parent for a subnet, you can use the `checkpoint list` command provided by the agent as follows:
```
jsoares, line 79:
```suggestion
Subnets are periodically committing checkpoints to their parent every `bottomup-check-period` (parameter defined when creating the subnet). If you want to inspect the information of a range of bottom-up checkpoints committed in the parent for a subnet, you can use the `checkpoint list` command provided by the agent as follows: 
```

jsoares: We could consider renaming this `bottomup-list`. Very minor but might be confusing.

docs/usage.md
```diff
@@ -37,28 +40,75 @@ $ ./bin/ipc-agent subnet send-value --subnet <subnet-id> --to <to-addr> <value>
...
+$ ./bin/ipc-agent cross-msg fund --subnet=/root/t01002 100
+
 ```
-Leaving a subnet will release the collateral for the validator and remove all the validation rights from its account. This means that if you have a validator running in that subnet, its validation process will immediately terminate.
+This command includes the cross-net message into the next top-down checkpoint after the current epoch. Once the top-down checkpoint is committed, you should see the funds in your account of the child subnet.
+
+>💡 Top-down checkpoints are not used to anchor the security of the parent into the child (as is the case for bottom-up checkpoints). They just include information of the top-down messages that need to be executed in the child subnet, and are a way for validators in the subnet to reach consensus on the finality on their parent.
+
+### Release
+In order to release funds from a subnet, your account must hold enough funds inside it. Releasing funds to the parent subnet can be permformed with the following comand:
```
jsoares, line 67: So the command works as expected but if you try to release in excess of the supply, you get:
> [2023-04-19T00:02:18Z ERROR ipc_agent::manager::bottomup] cannot submit checkpoint due to json_rpc error: {"code":1,"message":"GasEstimateMessageGas error: message execution failed: exit 20, reason: message failed with backtrace:\n00: f01002 (method 1487690799) -- send to t064 method 1468071864 aborted with code 20 (20)\n01: f064 (method 1468071864) -- error releasing circulating supply: wtf! we can't release funds below circ, supply. something went really wrong (20)\n (RetCode=20)"}

This might be a bit too alarming -- it's not that something went really wrong, just a user made a wrong input.

jsoares (approved):
Apologies as I started fixing multiple things I bumped into while testing this and went a bit outside of the scope. Anyhow, simplified instructions for cross-msgs and beyond, and fixed a few bugs. 

Beyond docs, note change to config/mod.rs, which was a bit misleading: our docs use 1234 and the API port and 1234 as the validator port, but the config defaulted to 1234 for the API, which might confuse users. This is more in line with our tutorials.

docs/usage.md
jsoares: Wait, I might have misread it. In fact, per https://github.com/consensus-shipyard/lotus/pull/161, the subnet account is initialised with 5000 FIL at genesis but this isn't counted towards the circulating supply because it didn't go through `fund`. So then we end up with the `release` returning success (fund check clears) but the message then failing validation on account of the circulating supply. Not great 🤔

docs/subnet.md
adlrocha: Good point. Clarified in the cli help.

docs/subnet.md
adlrocha: Thanks, they should only be here. So this should fix it.

docs/usage.md
adlrocha: I relaxed a bit the tone of that error messages (just in case someone sees it in the future), and since the fix of that PR, it should be hard for validators to encounter it as we are removing all genesis funds (except for the faucet).

docs/usage.md
adlrocha: I already changed in the codebase but was waiting to see if someone complaint, you did so I changed it also in cli and will push the fix in a minute :)

src/cli/commands/checkpoint/mod.rs
```diff
@@ -22,14 +22,14 @@ pub(crate) struct CheckpointCommandsArgs {
 impl CheckpointCommandsArgs {
     pub async fn handle(&self, global: &GlobalArguments) -> anyhow::Result<()> {
         match &self.command {
-            Commands::List(args) => ListBottomUpCheckpoints::handle(global, args).await,
+            Commands::BottomupList(args) => ListBottomUpCheckpoints::handle(global, args).await,
```
jsoares, line 25:
```suggestion
            Commands::ListBottomup(args) => ListBottomUpCheckpoints::handle(global, args).await,
```

src/cli/commands/checkpoint/mod.rs
```diff
@@ -22,14 +22,14 @@ pub(crate) struct CheckpointCommandsArgs {
...
             Commands::LastTopdown(args) => LastTopDownExec::handle(global, args).await,
         }
     }
 }
 
 #[derive(Debug, Subcommand)]
 pub(crate) enum Commands {
-    List(ListBottomUpCheckpointsArgs),
+    BottomupList(ListBottomUpCheckpointsArgs),
```
jsoares, line 33:
```suggestion
    ListBottomup(ListBottomUpCheckpointsArgs),
```

docs/usage.md
```diff
@@ -37,28 +47,75 @@ $ ./bin/ipc-agent subnet send-value --subnet <subnet-id> --to <to-addr> <value>
...
-### Listing checkpoints from a subnet
 
-Subnets are periodically committing checkpoints to their parent every `check-period` (parameter defined when creating the subnet). If you want to inspect the information of a range of checkpoints committed in the parent for a subnet, you can use the `checkpoint list` command provided by the agent as follows: 
+## Listing checkpoints from a subnet
+
+Subnets are periodically committing checkpoints to their parent every `bottomup-check-period` (parameter defined when creating the subnet). If you want to inspect the information of a range of bottom-up checkpoints committed in the parent for a subnet, you can use the `checkpoint list` command provided by the agent as follows: 
 ```bash
 # List checkpoints between two epochs for a subnet
-$ ./bin/ipc-agent checkpoint list --from-epoch <range-start> --to-epoch <range-end> --subnet <subnet-id>
+$ ./bin/ipc-agent checkpoint bottomup-list --from-epoch <range-start> --to-epoch <range-end> --subnet <subnet-id>
```
jsoares, line 89:
```suggestion
$ ./bin/ipc-agent checkpoint list-bottomup --from-epoch <range-start> --to-epoch <range-end> --subnet <subnet-id>
```

docs/usage.md
```diff
@@ -37,28 +47,75 @@ $ ./bin/ipc-agent subnet send-value --subnet <subnet-id> --to <to-addr> <value>
...
 
 # Example execution
-$ ./bin/ipc-agent checkpoint list --from-epoch 0 --to-epoch 100 --subnet root/t01002
+$ ./bin/ipc-agent checkpoint bottomup-list --from-epoch 0 --to-epoch 100 --subnet /root/t01002
```
jsoares, line 92:
```suggestion
$ ./bin/ipc-agent checkpoint list-bottomup --from-epoch 0 --to-epoch 100 --subnet /root/t01002
```

docs/usage.md
```diff
@@ -37,28 +47,75 @@ $ ./bin/ipc-agent subnet send-value --subnet <subnet-id> --to <to-addr> <value>
...
+$ ./bin/ipc-agent cross-msg release --subnet=/root/t01002 100
+```
+This command includes the cross-net message into a bottom-up checkpoint after the current epoch. Once the bottom-up checkpoint is committed, you should see the funds in your account in the parent. 
 
-### Listing checkpoints from a subnet
 
-Subnets are periodically committing checkpoints to their parent every `check-period` (parameter defined when creating the subnet). If you want to inspect the information of a range of checkpoints committed in the parent for a subnet, you can use the `checkpoint list` command provided by the agent as follows: 
+## Listing checkpoints from a subnet
+
+Subnets are periodically committing checkpoints to their parent every `bottomup-check-period` (parameter defined when creating the subnet). If you want to inspect the information of a range of bottom-up checkpoints committed in the parent for a subnet, you can use the `checkpoint list` command provided by the agent as follows:
```
jsoares, line 86:
```suggestion
Subnets are periodically committing checkpoints to their parent every `bottomup-check-period` (parameter defined when creating the subnet). If you want to inspect the information of a range of bottom-up checkpoints committed in the parent for a subnet, you can use the `checkpoint list-bottomup` command provided by the agent as follows: 
```

jsoares (approved):
For external consistency in the interface (last-topdown, list-bottomup, but also things like send-value), I think we should reverse the order here. I don't feel very strongly about it but it's my final suggestion -- everything looks and tests good otherwise.

PR #168, opened 2023-04-19
Wallet list

cryptoAtwill:
Add wallet list rpc call and cli handler.


adlrocha (changes requested):
We may have had a misunderstanding, the goal of this command is to return the list of wallets in the subnet handled by the peer, along with their balances.

src/cli/commands/wallet/list.rs
```diff
@@ -0,0 +1,51 @@
...
+        };
+
+        let addrs = json_rpc_client
+            .request::<WalletListResponse>(
+                json_rpc_methods::WALLET_LIST,
+                serde_json::to_value(params)?,
+            )
+            .await?;
+
+        log::info!("wallets in subnet {:} are {:?}", arguments.subnet, addrs);
```
adlrocha, line 38: We are still not showing the balance for each address here, right? I would love to include this in the docs (I think it could really improve the UX)

src/manager/lotus.rs
```diff
@@ -306,6 +306,21 @@ impl<T: JsonRpcClient + Send + Sync> SubnetManager for LotusSubnetManager<T> {
...
+        log::info!("list wallet in subnet");
+        self.lotus_client
+            .wallet_list()
+            .await?
+            .iter()
+            .map(|raw| Address::from_str(raw).map_err(|_| anyhow!("invalid addr: {raw:}")))
+            .collect::<Result<_>>()
+    }
+
+    async fn wallet_balance(&self, address: &Address) -> Result<TokenAmount> {
```
adlrocha, line 319: IIRC, `Filecoin.WalletList` already returns the balance for the list of addresses, I think is good to keep this `wallet_balance` RPC method here as we may need it in the future, but maybe you can save a few calls by getting all the info you need from wallet list.

adlrocha (approved):
We should update the CLI also to report the balances. Once this is done feel free to merge.

PR #171, opened 2023-04-20
Replace nohup suggestions

jsoares:
Sorry, got pulled into something else and forgot to PR this.


jsoares:
Merged without review as this is strictly better than the current broken status, but please say something if you see something.

PR #175, opened 2023-04-24
Refactor checkpoint submission

cryptoAtwill:
This is a proposed solution to address #172. 

Merge the implementation of bottom up and top down checkpoint submission to reduce duplicated code. This is just the trait definition. Once this is confirmed will migrate the detailed implementation.

Some key changes:
- Use one tokio thread to manage all subnets and validators. But use tokio with batch processing to submit checkpoint for each validator, see `CheckpointDriver` struct.
- Abstract both top-down and bottom-up submission into one common process while each can have its own implementation, see `CheckpointManager` trait.
- Submission of checkpoint to node mempool is encapsulated into its own trait. We might have different ways of submission, i.e. submit with no wait, submit with wait of timeout, submit with one epoch confirmation with timeout and etc. See `SubmissionStrategy` trait.


src/manager/checkpoint/manager.rs
```diff
@@ -0,0 +1,62 @@
...
+    fn checkpoint_type(&self) -> CheckpointType;
+
+    /// The checkpoint period to submit
+    fn checkpoint_period(&self) -> ChainEpoch;
+
+    /// The submit to memory pool strategy, i.e. determines if the submission to memory pool will
+    /// wait for the execution to complete or wait for a specific timeout.
+    fn submission_strategy(&self) -> &Self::SubmissionStrategy;
+
+    async fn sync_checkpoint_period(&self) -> Result<()>;
```
adlrocha, line 44: What is the purpose of this function?

src/manager/checkpoint/submit/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+use fvm_shared::address::Address;
+use fvm_shared::clock::ChainEpoch;
+use ipc_sdk::subnet_id::SubnetID;
+use serde::Serialize;
+
+/// The submission strategy. Depending on the speed requirement, the strategy to submit the checkpoint
+/// can vary. For example, we can have checkpoint submission without waiting for the message to be
+/// executed, or we can have checkpoint submission that waits for a specific timeout period.
+#[async_trait]
+pub trait SubmissionStrategy {
```
adlrocha, line 16: I would probably remove this abstraction for now, and have `submit_checkpoint` as part of the `CheckpointManager`, and it checkpoint manager can implement the checkpoint submission strategy at will. I don't expect an agent implementation to require more than one.

src/manager/checkpoint/subsystem.rs
```diff
@@ -0,0 +1,202 @@
...
+            // }) => Ok::<(), anyhow::Error>(()),
+            _ = subsys.on_shutdown_requested() => {
+                log::info!("Shutting down checkpointing subsystem");
+                Ok(())
+            }
+        }
+    }
+}
+
+struct CheckpointDriver {
```
adlrocha, line 43: I don't get why need so many abstractions, they are super confusing: 
CheckpointSubsystem --> CheckpointDriver --> CheckpointManagerWrapper --> Tokio tasks

Can't we simplify it a bit by just having a CheckpointSubsystems that spawns the CheckpointManagers for each of the checkpoints and that's it?

adlrocha (changes requested):
The current architecture is a bit confusing. We should probably draw a code architecture design really quick before starting with the code. We can sync at some point to discuss this.

src/manager/checkpoint/manager/bottomup.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    fn child_subnet(&self) -> &SubnetID {
+        todo!()
+    }
+
+    fn checkpoint_period(&self) -> ChainEpoch {
+        todo!()
+    }
+
+    async fn sync_checkpoint_period(&self) -> anyhow::Result<()> {
```
adlrocha, line 26: As we mentioned sync, we can probably remove this one and add an asynchronous constructor that gets the checkpoint period? Or maybe `checkpoint_period` could be async and the implementation could try to get the checkpoint period from an attribute from the manager and it if is not set it tries to fetch it from the state of the blockchain (I think I like the latter more)

src/manager/checkpoint/manager/bottomup.rs
```diff
@@ -0,0 +1,49 @@
...
+        todo!()
+    }
+
+    async fn obtain_validators(&self) -> anyhow::Result<Vec<Address>> {
```
adlrocha, line 30: Maybe rename? 
```suggestion
    async fn child_validators(&self) -> anyhow::Result<Vec<Address>> {
```

src/manager/checkpoint/manager/bottomup.rs
```diff
@@ -0,0 +1,49 @@
...
+        todo!()
+    }
+
+    async fn submit_checkpoint(
+        &self,
+        _epoch: ChainEpoch,
+        _previous_epoch: ChainEpoch,
```
adlrocha, line 37: Do we need the previous epoch for something? (maybe we do, without looking at the code it is not clear to me)

src/manager/checkpoint/subsystem.rs
```diff
@@ -1,66 +1,61 @@
...
-            // Each manager might have to handle multiple validators, we group them into
-            // (Manager, Validator) tuple pair task, so that we can fire the submission in batches
-            let tasks = break_into_tasks(managers).await?;
-
-            process_tasks(&tasks, &mut config_chan).await?;
+            let (top_down_managers, bottom_up_managers) =
+                setup_managers_from_config(&config.subnets)?;
+
+            loop {
+                select! {
```
adlrocha, line 38: Are the below comments still the case?

adlrocha (commented):
This is way cleaner, thank you Will. I think we can move on to the implementation.

adlrocha:
Closing in favor of #182 #184 and #185

src/manager/checkpoint/manager/bottomup.rs
cryptoAtwill: method removed already

src/manager/checkpoint/subsystem.rs
```diff
@@ -73,67 +68,27 @@ fn handle_response(response: anyhow::Result<()>) {
...
-        let validators = p.setup().await?;
+async fn process_managers<T: CheckpointManager>(managers: &[T]) -> anyhow::Result<()> {
+    // Tracks the start time of the processing, will use this to determine should sleep
+    let start_time = Instant::now();
 
-        let p = Arc::new(p);
-        for validator in validators {
-            pairs.push((p.clone(), validator));
-        }
+    // A loop that drives stream to the end
```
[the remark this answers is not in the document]
cryptoAtwill: this code is also removed.

src/manager/checkpoint/manager/bottomup.rs
cryptoAtwill: removed already

src/manager/checkpoint/subsystem.rs
cryptoAtwill: those comments are removed already

src/manager/checkpoint/manager.rs
cryptoAtwill: removed this function already

src/manager/checkpoint/submit/mod.rs
cryptoAtwill: remove trait already

src/manager/checkpoint/subsystem.rs
cryptoAtwill: already removed

adlrocha (approved):
I think this one LGTM, it has been reviewed in the previous PRs. Once we get #199 merged let's get this one merged.

PR #182, opened 2023-04-28
migrate bottom up checkpoint

cryptoAtwill:
Migrate bottom up checkpoint submission to the new code structure.


src/lotus/client.rs
```diff
@@ -59,6 +61,7 @@ mod methods {
     pub const IPC_LIST_BOTTOMUP_CHECKPOINTS: &str = "Filecoin.IPCListCheckpointsSerialized";
     pub const IPC_GET_TOPDOWN_MESSAGES: &str = "Filecoin.IPCGetTopDownMsgsSerialized";
     pub const IPC_GENESIS_EPOCH_FOR_SUBNET: &str = "Filecoin.IPCGetGenesisEpochForSubnet";
+    pub const IPC_HAS_VOTED_BOTTOM_UP_CHECKPOINT: &str = "Filecoin.IPCHasVotedBottomUpCheckpoint";
```
adlrocha, line 64: Wasn't this endpoint already implemented? (I may be wrong)

src/lotus/mod.rs
```diff
@@ -81,6 +81,9 @@ pub trait LotusClient {
     /// See: https://lotus.filecoin.io/reference/lotus/chain/#chainhead
     async fn chain_head(&self) -> Result<ChainHeadResponse>;
 
+    /// Returns the latest epoch of the chain
```
adlrocha, line 84:
```suggestion
    /// Returns the heaviest epoch for the chain
```

src/lotus/client.rs
```diff
@@ -489,6 +496,24 @@ impl<T: JsonRpcClient + Send + Sync> LotusClient for LotusJsonRPCClient<T> {
 
         Ok(checkpoints)
     }
+
+    /// Checks if the validator has submitted the
```
adlrocha, line 500: We are not using comments in the implementations, only in the trait.
```suggestion
```

src/lotus/client.rs
```diff
@@ -34,6 +34,8 @@ use crate::lotus::message::CIDMap;
 use crate::lotus::{LotusClient, NetworkVersion};
 use crate::manager::SubnetInfo;
 
+pub type DefaultLotusJsonRPCClient = LotusJsonRPCClient<JsonRpcClientImpl>;
```
adlrocha, line 37: Why are you setting this alias?

src/manager/checkpoint/manager/bottomup.rs
```diff
@@ -1,49 +1,288 @@
...
+    async fn last_executed_epoch(&self) -> anyhow::Result<ChainEpoch> {
+        let parent_head = self.parent_client.chain_head().await?;
+
+        // A key assumption we make now is that each block has exactly one tip set. We panic
+        // if this is not the case as it violates our assumption.
+        // TODO: update this logic once the assumption changes (i.e., mainnet)
+        assert_eq!(parent_head.cids.len(), 1);
+
+        let cid_map = parent_head.cids.first().unwrap().clone();
+        let parent_tip_set = Cid::try_from(cid_map)?;
```
adlrocha, line 71: Should we make the `parent_tipset` an argument for this one? That way we just need to get it one and we can reuse it in the rest of the functions.

src/manager/checkpoint/manager/bottomup.rs
```diff
@@ -1,49 +1,288 @@
...
+        Ok(subnet_actor_state
+            .bottom_up_checkpoint_voting
+            .last_voting_executed)
+    }
+}
 
 #[async_trait]
 impl CheckpointManager for BottomUpCheckpointManager {
-    type Checkpoint = BottomUpCheckpoint;
+    async fn obtain_validators(&self) -> anyhow::Result<Vec<Address>> {
```
adlrocha, line 86: Same here, we should maybe add the tipset as an argument and that way we can get the tipset once and get the data that we need from it.

src/manager/checkpoint/manager/bottomup.rs
```diff
@@ -1,49 +1,288 @@
...
+async fn obtain_checkpoint_period(
+    subnet_id: &SubnetID,
+    parent_client: &DefaultLotusJsonRPCClient,
+) -> anyhow::Result<ChainEpoch> {
+    log::debug!("Getting the bottom up checkpoint period for subnet: {subnet_id:?}");
+
+    // Read the parent's chain head and obtain the tip set CID.
+    let parent_head = parent_client.chain_head().await?;
+    let cid_map = parent_head.cids.first().unwrap().clone();
+    let parent_tip_set = Cid::try_from(cid_map)?;
```
adlrocha, line 275: Same here, is it worth adding the tipset as an argument?

adlrocha (approved):
LGTM. A few comments around re-using tipsets so we don't have to fetch them twice. Have you tested if it works, BTW? Once fixed feel free to merge

src/lotus/client.rs
cryptoAtwill: It's in the checkpoint, we need to specify the client lotus type, with alias, it's just much easier, otherwise every time, we need to write `LotusJsonRPCClient<JsonRpcClientImpl>`, it's a lot verbose and not intuitive as compared to `DefaultLotusJsonRPCClient`

src/manager/checkpoint/manager/bottomup.rs
cryptoAtwill: I have placed the obtain validators into a common function and taken it out of the trait. Also placed the latest executed method into the trait so that all validators can share the same parameter during each epoch submission.

src/manager/checkpoint/manager/bottomup.rs
cryptoAtwill: see comment above

src/manager/checkpoint/manager/bottomup.rs
cryptoAtwill: see comment above

src/lotus/client.rs
cryptoAtwill: yeah, removed already

src/manager/checkpoint/manager/bottomup.rs
```diff
@@ -1,49 +1,257 @@
...
 
     fn checkpoint_period(&self) -> ChainEpoch {
-        todo!()
+        self.checkpoint_period
     }
 
-    async fn sync_checkpoint_period(&self) -> anyhow::Result<()> {
-        todo!()
+    async fn last_executed_epoch(&self) -> anyhow::Result<ChainEpoch> {
+        let parent_head = self.parent_client.chain_head().await?;
```
adlrocha, line 84: You can probably `get parent head directly here and replace it for lines 84-92

adlrocha (commented):
A new round of comments that are really similar to the ones I already gave in the other PRs,  but just in case you forget to address these.

PR #184, opened 2023-05-02
Migrate top down

cryptoAtwill:
The migration of top down checkpoint submission to the new code structure.

TODO: might need to look into cache some parameters to reduce duplicated on chain data pull.


src/manager/checkpoint/manager/topdown.rs
```diff
@@ -1,19 +1,90 @@
...
+        write!(
+            f,
+            "top-down, parent: {:}, child: {:}",
+            self.parent.id, self.child_subnet.id
+        )
+    }
+}
+
+impl TopDownCheckpointManager {
+    async fn sync_subnet_gateway_state(&self) -> anyhow::Result<IPCReadGatewayStateResponse> {
```
adlrocha, line 59: Proposal of renaming, why we need the `sync_subnet` prefix for the fn?
```suggestion
    async fn child_gateway_state(&self) -> anyhow::Result<IPCReadGatewayStateResponse> {
```

src/manager/checkpoint/manager/topdown.rs
```diff
@@ -1,19 +1,90 @@
...
+        let child_head = self.child_client.chain_head().await?;
+        let cid_map = child_head.cids.first().unwrap().clone();
+        let child_tip_set = Cid::try_from(cid_map)?;
+
+        self.child_client
+            .ipc_read_gateway_state(child_tip_set)
+            .await
+    }
+
+    async fn parent_tipset(&self) -> anyhow::Result<Cid> {
```
adlrocha, line 69: This function doesn't return the tipset but its key, what about renaming it to:
```suggestion
    async fn parent_head(&self) -> anyhow::Result<Cid> {
```

src/manager/checkpoint/manager/topdown.rs
```diff
@@ -1,19 +1,90 @@
...
+        let parent_head = self.parent_client.chain_head().await?;
+        let cid_map = parent_head.cids.first().unwrap().clone();
+        Cid::try_from(cid_map)
+    }
+
+    async fn child_tipset(&self) -> anyhow::Result<Cid> {
```
adlrocha, line 75:
```suggestion
    async fn child_head(&self) -> anyhow::Result<Cid> {
```

src/manager/checkpoint/manager/topdown.rs
```diff
@@ -22,47 +93,135 @@ impl CheckpointManager for TopDownCheckpointManager {
...
+            "checkpoint at epoch {:} for manager: {:} published with cid: {:?}, executed",
+            epoch,
+            self,
+            message_cid,
+        );
+
+        Ok(())
     }
 
     async fn should_submit_in_epoch(
```
adlrocha, line 189: I don't know if this also exists for the bottom-up checkpoint. If this is not the case we should also check it here. Again leaving the comment before reviewing the corresponding PR so I don't forget :)

src/manager/checkpoint/manager/topdown.rs
```diff
@@ -22,47 +93,135 @@ impl CheckpointManager for TopDownCheckpointManager {
...
     }
 
     async fn presubmission_check(&self) -> anyhow::Result<bool> {
-        Ok(true)
+        let state = self.sync_subnet_gateway_state().await?;
+        Ok(state.initialized)
     }
 }
+
+async fn obtain_checkpoint_period(
```
adlrocha, line 211: `obtain` sounds a bit weird, what about `get`, or `pull`, in case you want to make more explicit that we are pulling it from the chain state.
```suggestion
async fn get_checkpoint_period(
```

adlrocha (commented):
LGTM. Check my comments before merging, but nothing critical, feel free to merge after you address them.

> TODO: might need to look into cache some parameters to reduce duplicated on chain data pull.

Regarding the above, would you mind adding an issue with the things you think we could cache to minimize the chain interactions.

src/manager/checkpoint/manager/topdown.rs
cryptoAtwill: updated

src/manager/checkpoint/manager/topdown.rs
```diff
@@ -22,47 +93,135 @@ impl CheckpointManager for TopDownCheckpointManager {
...
+    subnet_id: &SubnetID,
+    child_client: &DefaultLotusJsonRPCClient,
+) -> anyhow::Result<ChainEpoch> {
+    log::debug!("Getting the top down checkpoint period for subnet: {subnet_id:?}");
+
+    // Read the child's chain head and obtain the tip set CID.
+    log::debug!("Getting child tipset and starting top-down checkpointing manager");
```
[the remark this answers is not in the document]
cryptoAtwill: updated

PR #185, opened 2023-05-03
Integrate new checkpoint

cryptoAtwill:
Contains fixes to errors during integration testing.


src/manager/checkpoint/bottomup.rs
adlrocha, line 1: Should we remove this file all along? It looks like this PR removes all the code.

src/manager/checkpoint/subsystem.rs
adlrocha, line 1: Should we move this to `mod.rs` (which is empty) instead of having it in an independent file?

src/manager/checkpoint/subsystem.rs
```diff
@@ -179,12 +188,16 @@ async fn submit_till_current_epoch(manager: &impl CheckpointManager) -> Result<(
...
+        current_epoch,
+        SUBMISSION_LOOK_AHEAD_EPOCH + last_executed_epoch,
+    );
+    while next_epoch < cut_off_epoch {
         // now we process each validator
         for validator in &validators {
             log::debug!("submit checkpoint for validator: {validator:?} in manager: {manager:}");
 
-            if manager
+            if !manager
```
adlrocha, line 200: Would you mind explaining this and the `cut_off_epoch` (I may understand it better in the next PR, but just in case I forget).

adlrocha (approved):
I added a few comments here and there with things to be improved, but once these are fixed feel free to merge :pray:

src/manager/checkpoint/bottomup.rs
cryptoAtwill: it's removed now

src/manager/checkpoint/subsystem.rs
cryptoAtwill: I put a comment in the PR as well, just paste it here for reference:
```
Instead of loop all the way to `current_epoch`, we loop till `cut_off_epoch`.
Reason because if the current epoch is significantly greater than last_executed_epoch and there are lots of validators in the network, loop all the way to current epoch might have some outdated data. Set a cut off epoch such that validators can sync with chain more regularly.
```

PR #187, opened 2023-05-04
ID-1: Migrate and adapt `forest` key management crates 

adlrocha:
This is the first PR in the effort of implementing identity and key management in the agent so there is no need to import/export keys directly on subnet peer implementations. That way the agent will be able to sign and send message to a subnet without having to import a key on the peer they are connected to.

This PR migrates and adapts forest's key management code so it can be used as a base for the agent.


cryptoAtwill:
I feel it might be better to move these code into a separate crate in the repo and import in agent?

cryptoAtwill (approved):

identity/src/wallet.rs
```diff
@@ -0,0 +1,466 @@
...
+        for item in key_vec.into_iter() {
+            keys.insert(item.address, item);
+        }
+        Wallet { keys, keystore }
+    }
+
+    // If this key does not exist in the keys hashmap, check if this key is in
+    // the keystore, if it is, then add it to keys, otherwise return Error
+    /// Return the key that is resolved by a given address,
+    pub fn find_key(&mut self, addr: &Address) -> Result<Key, Error> {
```
cryptoAtwill, line 72: This api is causing some issues downstream. From the function name, it's a `getter`, but it's taking a `&mut self`, this makes anyone who wants to find a key that holds the `wallet` need to use `&mut self`. And it complicates things downstream. If possible, we should remove the `&mut`.

PR #189, opened 2023-05-05
Identity API integration

adlrocha:
Depends on https://github.com/consensus-shipyard/ipc-agent/pull/187

This PR integrates the identity management into the JSON RPC server.


src/server/handlers/wallet/balances.rs
```diff
@@ -46,7 +48,11 @@ impl JsonRPCRequestHandler for WalletListHandler {
         };
 
         let manager = conn.manager();
-        let addresses = manager.wallet_list().await?;
+
+        let addresses = self.wallet.read().unwrap().list_addrs()?;
+        // Create a new Arc for wallet so it is pulled in the async block
+        // from below.
+        let _arc_wallet = Arc::clone(&self.wallet);
```
cryptoAtwill, line 55: do we still need this if `_arc_wallet` is not used at all?

src/server/handlers/wallet/export.rs
```diff
@@ -0,0 +1,49 @@
...
+use crate::identity::Wallet;
+use crate::server::JsonRPCRequestHandler;
+use async_trait::async_trait;
+use fvm_shared::address::Address;
+use serde::{Deserialize, Serialize};
+use std::str::FromStr;
+use std::sync::{Arc, RwLock};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct WalletExportParams {
```
cryptoAtwill, line 15: not sure if this will pose some security risks as it's exposed over json rpc

cryptoAtwill (approved):

src/server/handlers/wallet/balances.rs
adlrocha: Unfortunately, we do because we use an `async_move` in the next line and if we don´t do this the wallet won't implement `Send` and the compiler won't be happy. This makes it happy :)

src/server/handlers/wallet/export.rs
adlrocha: Good point, is definitely an issue without proper authentication. I will probably make it available only through the CLI in the next PR, so only someone with access to the local process can export the keys. Thanks!

PR #190, opened 2023-05-08
Replace u64 with f64 for FIL amount

cryptoAtwill:
This is a PR to address the issue #178. Uses `f64` over `u64` for amount in send value calls.


adlrocha (approved):
really nice. LGTM. Let's merge :)

PR #196, opened 2023-05-09
Sign transaction in agent

cryptoAtwill:
Support signing mpool messages in agent, added the following methods:
1. Estimate gas.
2. Obtain nonce.
3. Sign the memory pool message using the target wallet.

@adlrocha currently the files changed contains the latest main code, might need to pull latest main into `adlrocha/id-2`.


src/lotus/client.rs
```diff
@@ -491,6 +537,100 @@ impl<T: JsonRpcClient + Send + Sync> LotusClient for LotusJsonRPCClient<T> {
     }
 }
 
+impl<T: JsonRpcClient + Send + Sync> LotusJsonRPCClient<T> {
+    fn sign_mpool_message(&self, msg: &MpoolPushMessage) -> anyhow::Result<Signature> {
+        if self.wallet_store.is_none() {
+            return Err(anyhow!("key store not set, function not supported"));
+        }
+
+        let message = fvm_shared::message::Message {
```
adlrocha, line 546: Out of a curiosity, is this how forest does it? By creating an `fvm::Message` and signing its CID?

src/lotus/client.rs
```diff
@@ -500,4 +640,105 @@ impl LotusJsonRPCClient<JsonRpcClientImpl> {
...
+            },
+            "Signature": {
+                "Type": sig_type as u8,
+                "Data": sig_encoded,
+            }
+        }
+    ])
 }
+
+// #[cfg(test)]
```
adlrocha, line 704: Should we remove this one or do you want to keep it for reference in some form? If this is the case we should consider making it cleaner.

src/lotus/mod.rs
```diff
@@ -43,6 +43,11 @@ pub trait LotusClient {
         msg: MpoolPushMessage,
     ) -> Result<MpoolPushMessageResponseInner>;
 
+    /// Push the unsigned message to memory pool. This will ask the local key store to sign the message.
+    /// In this case, make sure `from` is actually present in the local key store.
```
adlrocha, line 47: How are we handling the key not being present in the local wallet? We should just abort, but just wondering in case we are sharing enough feedback about the error to users.

adlrocha (approved):
LGTM, amazing job! A few comments maybe worth addressing, but feel free to merge when ready. 

Once this is merged I think we can directly merge `adlrocha/id-1` to main, we are not introducing any breaking changes so far.

src/lotus/client.rs
cryptoAtwill: yeah, I copied from them. They are actually using `message.cid()` with `Impl Cbor ...`. But `Cbor` trait seems to be deprecated. So I just copied the implementation over.

src/lotus/client.rs
cryptoAtwill: Removed

src/lotus/mod.rs
cryptoAtwill: This one seems need to be handled from `WalletStore`. Error message will be key not exists.
image

adlrocha: That works, we can keep it as is for now. Thanks!

src/lotus/client.rs
adlrocha: Cool!

Issue #186, opened 2023-05-04
Checkpointing: Add state root in bottom-up checkpoint


adlrocha:
The proof for bottom-up checkpoints is currently the key of the tipset for the epoch of the checkpoint being submitted. While this allows to follow that the history of the chain is consistent, it is a quite weak proof. To make it more robust and start moving towards a way to spawn a subnet from the state of a checkpoint, we should include in the proof the state root in the checkpoint's epoch, or both: the tipset key and the state root. The proof could be the serialization of:

```rust
struct {
  tipset: Cid
  state: Cid
}
```

cryptoAtwill:
btw, @adlrocha I think it's also good to add a version as the first field so that it's more upgrading friendly, something like:
```
enum CheckProof {
  V1 { tipset: Cid, state: Cid }
}
```

adlrocha:
> btw, @adlrocha I think it's also good to add a version as the first field so that it's more upgrading friendly, something like:

LGTM. Let's do this, even if we make breaking changes for versions initially until we make it to mainnet. I am thinking that if we want these proofs to be custom for subnets we may need to add some `metadata` name or unique id in the future, but version should be fine for now.

adlrocha:
Implemented in #199

PR #199, opened 2023-05-17
Bottom up checkpoint proof

cryptoAtwill:
Support more complex bottom up proof as described in #186.


src/manager/checkpoint/proof/v1.rs
```diff
@@ -0,0 +1,31 @@
...
+use crate::lotus::LotusClient;
+use anyhow::anyhow;
+use cid::Cid;
+use fvm_shared::clock::ChainEpoch;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct V1Proof {
+    tip_set: Vec<Cid>,
+    states: Vec<Cid>,
```
adlrocha, line 12: Why is `states` a `Vec<Cid>`?

adlrocha: Oh, I see. You set the parent root for all the blocks in the tipset. We only have one block per tipset so I would suggest using `Cid` instead of `Vec<Cid>` and assigning the one for block 0 (the only one present). I would probably do the same also for the `tip_set` (although this one is not as critical, we can leave the `Vec<Cid>`.

src/manager/checkpoint/proof/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+//! The checkpoint proof structs
+
+mod v1;
+
+use crate::lotus::LotusClient;
+pub use crate::manager::checkpoint::proof::v1::V1Proof;
+use fvm_shared::clock::ChainEpoch;
+use serde::{Deserialize, Serialize};
+
+/// The different versions of checkpoint proof
```
adlrocha, line 12:
```suggestion
/// The different versions of checkpoint proofs supported
```

adlrocha (approved):
Just a slight change on how we construct the proof, but LGTM. If you have tested it and it works let's get this one merged :)

src/manager/checkpoint/proof/v1.rs
cryptoAtwill: updated

src/manager/checkpoint/proof/v1.rs
```diff
@@ -0,0 +1,37 @@
...
+        .await?;
+
+    let blocks = response.blocks_state_roots()?;
+    if blocks.is_empty() {
+        return Err(anyhow!("height {height:} has zero blocks"));
+    }
+
+    Ok(V1Proof {
+        tip_set: response.tip_set_cids()?,
+        state: blocks[0],
```
adlrocha, line 35:
```suggestion
        // we assume tipsets with a single block in subnets.
        // we can directly get the `parent_state` for the only block.
        state: blocks[0],
```

PR #201, opened 2023-05-18
Add auto complete

cryptoAtwill:
Add auto complete. Each shell might have different setup, for `zsh`, just do:
```
ipc-agent --cli-autocomplete-gen zsh > /usr/local/share/zsh/site-functions/_ipc-agent
compinit 

# if this does not work for you, try `compdef _ipc-agent ipc-agent`
```


adlrocha (approved):
This works for me. Thanks. 

I was going to address the doc changes here, but we should change something else first for it to work. I tracked it in this issue: https://github.com/consensus-shipyard/ipc-agent/issues/203

src/cli/commands/mod.rs
```diff
@@ -55,10 +58,13 @@ enum Commands {
 )]
 #[command(propagate_version = true)]
 struct IPCAgentCliCommands {
+    // If provided, outputs the completion file for given shell
+    #[arg(long = "generate", value_enum)]
```
adlrocha, line 62: Should we call it `install_autocompletion` or `autocompletion`? Generate seems like too general

adlrocha (changes requested):
A few things to consider: 
- Change the name of the autocompletion generator flag.
- Can we add a `Makefile` rule to install autocompletion, something like `make install-autocomplete`? 
- Would you mind updating the docs to describe how to install autocompletion? Maybe we can even have a `make install` rule that does that automatically for you. 

Finally, why are you having trouble with `bash`? Let me know if I can help in any way.

jsoares:
This warms my heart.

jsoares:
Are we still pursuing this or should we just close?

cryptoAtwill:
@jsoares Oh yeah, this one is pretty close to done. I will run some extra test now then should be good for review.

src/cli/commands/mod.rs
cryptoAtwill: I guess this is the convention from clap?

adlrocha (approved):
@cryptoAtwill, if you tested it after rebasing with `dev` then LGTM. Thank you for taking thsi one to the finish line :) (it got stuck for a while haha)

PR #202, opened 2023-05-18
Cli wallet import

cryptoAtwill:
Fix cli cannot import wallet.


adlrocha (approved):

adlrocha (approved):
This works for me. Thanks.

I was going to address the doc changes here, but we should change something else first for it to work. I tracked it in this issue: https://github.com/consensus-shipyard/ipc-agent/issues/203

PR #204, opened 2023-05-18
fix export and docs

adlrocha:
This PR: 
- Fixes wallet export so it reads from the keystore and uses the right format so it can be imported.
- It updates docs to use ipc-agent identity management.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -142,9 +142,9 @@ Although we set a minimum of 2 active validators in the previous, we'll deploy 3
 ```
 * Export each wallet (WALLET_1, WALLET_2, and WALLET_3) by substituting their addresses below
 ```bash
-./lotus/eudico wallet export --lotus-json <WALLET_1> > ~/.ipc-agent/wallet1.key
-./lotus/eudico wallet export --lotus-json <WALLET_2> > ~/.ipc-agent/wallet2.key
-./lotus/eudico wallet export --lotus-json <WALLET_3> > ~/.ipc-agent/wallet3.key
+./ipc-agent/bin/ipc-agent wallet export --address=<WALLET_1> --output=~/.ipc-agent/wallet1.key
+./ipc-agent/bin/ipc-agent wallet export --address=<WALLET_2> --output=~/.ipc-agent/wallet2.key
+./ipc-agent/bin/ipc-agent wallet export --address=<WALLET_3> --output=~/.ipc-agent/wallet3.key
 ```
 * We also need to fund the wallets with enough collateral to; we'll send the funds from our default wallet 
 ```bash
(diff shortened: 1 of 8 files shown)


cryptoAtwill (approved):

docs/quickstart.md
```diff
@@ -163,12 +178,7 @@ We can deploy the subnet nodes. Note that each node should be importing a differ
...
 ## Step 7: Configure the IPC agent
 
-For ease of use, we'll import the remaining keys into the first validator, via which the IPC Agent will act on behalf of all.
-
-* Copy the wallet keys into the docker container and import them
-```bash
-docker cp ~/.ipc-agent/wallet2.key <CONTAINER_NAME_1>:/input.key && docker exec -it <CONTAINER_NAME_1> eudico wallet import --format=json-lotus input.key
-docker cp ~/.ipc-agent/wallet3.key <CONTAINER_NAME_1>:/input.key && docker exec -it <CONTAINER_NAME_1> eudico wallet import --format=json-lotus input.key
+For ease of use, we'll import the remaining keys into the first validator, via which the IPC Agent will act on behalf of all. All the keys should already be imported in the wallet of the agent, if for some reason this is not the case you can import it through: `./ipc-agent/bin/ipc-agent wallet import --path=~/.ipc-agent/wallet.key`
 ```
```
jsoares, line 182:
```suggestion
For ease of use, we'll import the remaining keys into the first validator, via which the IPC Agent will act on behalf of all.
```

There was a stray code block opening here. I also don't think we need the import instructions here as you should never run into that scenario when following the instructions.

docs/subnet.md
```diff
@@ -10,7 +10,12 @@ We provide instructions for running both a [simple single-validator subnet](#run
 
 ### Exporting wallet keys
 
-In order to run a validator in a subnet, we'll need a set of keys to handle that validator. To export the validator key from a wallet that may live in another network into a file (like the wallet address we are using in the rootnet), we can use the following Lotus command:
+In order to run a validator in a subnet, we'll need a set of keys to handle that validator. To export the validator key from your agent you need to run: 
+```bash
+./ipc-agent/bin/ipc-agent wallet export --address=<address-to-export> --output=<output file>
+```
+
+To export a wallet that may live in another network into a file (like the wallet address we are using in the rootnet), we can use the following Lotus command:
```
jsoares, line 18: Just to be clear, the idea here is that you're providing instructions to export both from ipc-agent (if using identity management) or from raw eudico (for keys not held by the agent). Is that it?

docs/subnet.md
```diff
@@ -31,8 +36,12 @@ $ docker exec -it ipc_root_1234 eudico wallet export --lotus-json t1cp4q4lqsdhob
 ```
 
 ### Importing wallet keys
+Your agent handles the keys for all of your addresses in IPC and is responsible for signing the transactions to the different networks. To import a key to the agent you need can use: 
+```bash
+`./ipc-agent/bin/ipc-agent wallet import --path=<wallet-key-file-path>`
+```
 
-Depending on whether the subnet is running inside a docker container or not, you may need to import keys into a node. You may use the following commands to import a wallet to a subnet node: 
+Depending on whether the subnet is running inside a docker container or not, you may need to import keys directly into a node to run your validator. You may use the following commands to import a wallet to a subnet node:
```
jsoares, line 44: Could you ellaborate? I understand that the export methodology differs based on circumstances, but it's less obvious to me when you'd need to import into the node vs. agent.

jsoares (approved):

jsoares:
I haven't actually run the instructions, just reviews the text. Will do that on Monday as I need to deploy a new subnet anyway.

docs/subnet.md
adlrocha: Yes, that was the idea, just in case someone has a set of keys on a lotus peer that they want to use in the agent.

adlrocha: I rephrased the paragraph a bit to make a bit clearer. Thank you for raising it.

docs/subnet.md
adlrocha: Fixed.

adlrocha:
Comments addressed. I will wait to get it merged and to cut the release before the restart. Thanks :pray:

PR #205, opened 2023-05-19
Fix checkpoint config load

cryptoAtwill:
Fix checkpoint subsystem initialization fail on config load crash the program.


src/manager/checkpoint/mod.rs
```diff
@@ -56,8 +56,19 @@ impl IntoSubsystem<anyhow::Error> for CheckpointSubsystem {
...
-                setup_managers_from_config(&config.subnets, self.wallet_store.clone()).await?;
+            let (top_down_managers, bottom_up_managers) = match setup_managers_from_config(
+                &config.subnets,
+                self.wallet_store.clone(),
+            )
+            .await
+            {
+                Ok(r) => r,
+                Err(e) => {
+                    log::error!("Please check configuration! Cannot start the checkpoint subsystem due to config error: {e:}. Sleep 15 seconds before retry.");
```
adlrocha, line 67: Can we instead of sleep an retryh just not start and try again only after a config reload? Or you think it would be better to retry periodically?

src/manager/checkpoint/mod.rs
```diff
@@ -56,8 +56,19 @@ impl IntoSubsystem<anyhow::Error> for CheckpointSubsystem {
...
+                    sleep(Duration::from_secs(15)).await;
```
adlrocha, line 68: Also, can we make this a config parameter?

adlrocha (changes requested):
A few suggestions, let me know what you think

src/manager/checkpoint/mod.rs
cryptoAtwill: Yeah, this works, save the config parameter

PR #206, opened 2023-05-22
Integration testing infra

cryptoAtwill:
This PR setups the child subnet infra based on the requirement number of validators.

This is useful when testing subnet checkpoints where we need multiple validators to submit checkpoints to the parent or to the subnet itself.

In `/integration-testing/src/main.rs` contains the configuration for the subnet topology and binary path definition.
In `/integration-testing/src/infra` contains the process flow for spawn subnet nodes and their validators.

To test this, one must first manually start the root net and the ipc agent. Then start the process with:
```
cd integration-testing
cargo build --release
../target/release/integration-testing
```


integration-testing/src/infra/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+
+use ipc_sdk::subnet_id::SubnetID;
+use std::sync::atomic::{AtomicU16, Ordering};
+use std::sync::Arc;
+
+const DEFAULT_IPC_AGENT_URL: &str = "http://localhost:3030/json_rpc";
+const DEFAULT_NODE_API_BASE_PORT: u16 = 1230;
+const DEFAULT_MIN_STAKE: f64 = 1.0;
+
+pub struct SubnetTopology {
```
adlrocha, line 16: Do you mind adding comments here so it is clear what each field of the struct is?

integration-testing/src/infra/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub id: Option<SubnetID>,
+    pub name: String,
+    pub number_of_nodes: usize,
+    pub eudico_binary_path: String,
+    pub parent: SubnetID,
+    pub root_address: String,
```
adlrocha, line 22: This one is one of the addresses that is not clear to me

integration-testing/src/infra/util.rs
```diff
@@ -0,0 +1,165 @@
...
+    from: String,
+    subnet: String,
+    collateral: f64,
+    validator_net_addr: String,
+) -> anyhow::Result<()> {
+    let json_rpc_client = JsonRpcClientImpl::new(ipc_agent_url.parse()?, None);
+
+    // The json rpc server will handle directing the request to
+    // the correct parent.
+    let params = JoinSubnetParams {
```
adlrocha, line 57: Quick idea: this code looks really similar to the one we use in the agent's CLI commands. Should we de-duplicate this code by creating an agent `client` or `sdk` crate that includes these (the same you have `utils`) and that applications or external crates like this one can use?

integration-testing/src/infra/subnet.rs
```diff
@@ -0,0 +1,562 @@
...
+    log::info!("first node net addr: {:?}", first_node_addr);
+
+    for node in &nodes {
+        node.connect_peer(&first_node_addr)?;
+    }
+
+    Ok(nodes)
+}
+
+pub struct SubnetNode {
```
adlrocha, line 170: Should we move all node-specific code to a `node.rs` file and leave here subnet info or they are too related?

adlrocha: Also, would you mind commenting a bit the structs and functions of this file?

integration-testing/src/main.rs
```diff
@@ -0,0 +1,42 @@
...
+use std::str::FromStr;
+use std::sync::atomic::AtomicU16;
+use std::sync::Arc;
+use std::thread::sleep;
+use std::time::Duration;
+
+mod infra;
+
+#[tokio::main]
+async fn main() {
```
adlrocha, line 13: Would it make sense to make this into a CLI so we can pass the tests that we want to run, flags to config the infra deployment, etc.?

adlrocha (approved):
THIS IS AN AMAZING JOB!! I think is going to make our lives way easier. I left a few suggestions, would you mind having a look? Thanks!!

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = [".", "ipld/resolver", "testing/e2e", "identity"]
+members = [".", "ipld/resolver", "testing/e2e", "identity", "testing/child-subnet-infra"]
```
adlrocha, line 2: I would call it `itests` directly for integration tests.
```suggestion
members = [".", "ipld/resolver", "testing/e2e", "identity", "testing/itest"]
```

src/cli/commands/wallet/import.rs
```diff
@@ -16,7 +17,27 @@ use crate::jsonrpc::{JsonRpcClient, JsonRpcClientImpl};
 use crate::lotus::message::wallet::WalletKeyType;
 use crate::server::wallet::import::{WalletImportParams, WalletImportResponse};
 
-pub(crate) struct WalletImport;
+pub struct WalletImport;
```
adlrocha, line 20: What is the purpose of this change?

testing/child-subnet-infra/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "child-subnet-infra"
```
adlrocha, line 2:
```suggestion
name = "itest"
```

testing/child-subnet-infra/README.md
```diff
@@ -0,0 +1,14 @@
+# Child Subnet Infra
```
adlrocha, line 1: Depending on how you are planning to run the tests this README should explain how to run the test cases you are currently working on for the integration tests.

adlrocha (changes requested):
A few suggestions, and I see that some of the comments may have not been addressed.

adlrocha (approved):

PR #208, opened 2023-05-24
Add ipc agent sdk

cryptoAtwill:
Add ipc agent sdk for code deduplication.


src/sdk/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+pub use wallet::LotusJsonKeyType;
+
+use url::Url;
+
+/// The ipc agent client
+pub struct IpcAgentClient<T> {
+    json_rpc_client: T,
+}
+
+impl<T> IpcAgentClient<T> {
```
adlrocha, line 19: Should we bound a bit more the type so only Transports can be used here?

adlrocha (approved):
Amazing job. Just a slight comment in case you think it makes sense

src/sdk/mod.rs
cryptoAtwill: I feel in rust the convention is the `impl` would be a the place to bound it, the struct definition is more general. But I could be wrong, but this is the pattern in many repos I saw in rust.

PR #210, opened 2023-05-25
Checkpoint integration tests

cryptoAtwill:
Refactoring integration testing and ipc agent (slight modification in the sdk and config) so that we can perform more integration testing on the different validators setup:
- [x] update ipc agent config and trigger config reload automatically
- [x] break each spawn node and validator steps into smaller function call
- [x] support tear down nodes and validators, so that we can kill nodes randomly
- [x] add initial test cases for submitting checkpoints


src/config/mod.rs
```diff
@@ -81,4 +81,18 @@ impl Config {
         let contents = tokio::fs::read_to_string(path).await?;
         Config::from_toml_str(contents.as_str())
     }
+
+    pub async fn write_to_file_async(&self, path: impl AsRef<Path>) -> Result<()> {
```
adlrocha, line 85: Just a quick check: Why do we need this function?

adlrocha (approved):
LGTM. What does the "testing in progress" mean? 

Be sure to fix the CI before merging, please.

src/config/mod.rs
cryptoAtwill: yep yep

src/config/server.rs
```diff
@@ -1,11 +1,11 @@
 // Copyright 2022-2023 Protocol Labs
 // SPDX-License-Identifier: MIT
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use std::net::SocketAddr;
 
 pub const JSON_RPC_ENDPOINT: &str = "json_rpc";
 
-#[derive(Deserialize, Clone, Debug)]
+#[derive(Deserialize, Serialize, Clone, Debug)]
```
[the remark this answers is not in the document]
cryptoAtwill: In the integration testing..., because we have to write them to file.

testing/itest/src/main.rs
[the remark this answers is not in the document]
cryptoAtwill: yeah, I'm planning to write in the `bin`, to provide more flexibility to run various setups.

Makefile
```diff
@@ -6,7 +6,7 @@ build:
 	cargo build -Z unstable-options --release --out-dir ./bin
 
 test:
-	cargo test --release --workspace --exclude ipc_e2e
+	cargo test --release --workspace --exclude ipc_e2e itest
```
adlrocha, line 9: If we excluse `itest` let's add a new rule called `make itests` that runs integration tests. WDYT?

testing/itest/examples/no_create_subnet.rs
adlrocha, line 1: Why it's called `no_create_subnet`? Would you mind adding a note in the README of the itests when you mention the `example` scripts with what each example does? Thanks!

adlrocha (approved):
LGTM. A few comments and a q: do we have in the README instruction on how to deploy the underlying dependencies needed to run the tests? Mainly, lotus?

testing/itest/examples/no_create_subnet.rs
cryptoAtwill: I added the comments in the test example files, this way might be easier for dev as there is no need to look back to the README.

Makefile
cryptoAtwill: added

PR #213, opened 2023-05-30
Integrate Eth client and EVM-based subnets

adlrocha:
Superseded by: https://github.com/consensus-shipyard/ipc-agent/pull/221


src/manager/evm.rs
```diff
@@ -0,0 +1,185 @@
...
+use crate::lotus::message::mpool::MpoolPushMessage;
+use crate::lotus::message::state::StateWaitMsgResponse;
+use crate::lotus::message::wallet::WalletKeyType;
+use crate::lotus::LotusClient;
+
+use super::subnet::SubnetManager;
+
+type MiddlewareImpl = SignerMiddleware<Provider<Http>, Wallet<SigningKey>>;
+
+pub struct EthSubnetManager<M: Send + Sync> {
```
cryptoAtwill, line 38: I think we can exclude Send Sync in the struct as we are confining it in the impl

src/server/handlers/manager/subnet.rs
```diff
@@ -4,24 +4,26 @@
...
 use ipc_identity::Wallet;
 use ipc_sdk::subnet_id::SubnetID;
 use std::sync::{Arc, RwLock};
 
 /// The subnet manager connection that holds the subnet config and the manager instance.
-pub struct Connection<T: JsonRpcClient> {
+pub struct Connection {
     subnet: Subnet,
-    manager: LotusSubnetManager<T>,
+    manager: Box<dyn SubnetManager>,
```
cryptoAtwill, line 15: Why not pass SubnetManager as generic parameter but use Box + dyn?

cryptoAtwill (approved):

src/manager/evm.rs
adlrocha: Good point, fixed!

src/server/handlers/manager/subnet.rs
adlrocha: Oh, I forgot to push the version where I was using generics. That works, but good catch, thanks!

PR #215, opened 2023-05-31
Evm create join functions

cryptoAtwill:
A lot of the changes are related subnet id update and ROOTNET_ID removal. For review, please focus on the file: `src/manager/evm.rs`, `src/server/handlers/manager/create.rs` and `src/server/handlers/manager/join.rs`.


Cargo.toml
```diff
@@ -82,9 +83,11 @@ fvm_ipld_blockstore = "0.1"
 fvm_ipld_encoding = "0.3"
 fvm_shared = { version = "=3.0.0-alpha.17", default-features = false, features = ["crypto"] }
 fil_actors_runtime = { git = "https://github.com/consensus-shipyard/fvm-utils", features = ["fil-actor"] }
-ipc-sdk = { git = "https://github.com/consensus-shipyard/ipc-actors.git", tag = "v0.2.0" }
-ipc-subnet-actor = { git = "https://github.com/consensus-shipyard/ipc-actors.git", features = [], tag = "v0.2.0" }
-ipc-gateway = { git = "https://github.com/consensus-shipyard/ipc-actors.git", features = [],  tag = "v0.2.0"  }
+
```
adlrocha, line 86: You can stop pointing to this branches to avoid having to propagate the changes for the new implementation of the `SubnetID`, that way it will be easier to pick up the changes in the rebase.

src/server/handlers/manager/join.rs
```diff
@@ -52,7 +52,7 @@ impl JsonRPCRequestHandler for JoinSubnetHandler {
         let collateral = handlers::f64_to_token_amount(request.collateral)?;
 
         let subnet_config = conn.subnet();
-        check_subnet(subnet_config)?;
+        // check_subnet(subnet_config)?;
```
adlrocha, line 55: We should still check the parent (I guess you commented it due to the Subnet change)

src/manager/evm.rs
```diff
@@ -26,127 +30,213 @@ pub type MiddlewareImpl = SignerMiddleware<Provider<Http>, Wallet<SigningKey>>;
...
+            top_down_check_period: params.topdown_check_period as u64,
+            // TODO: update this variable properly
+            majority_percentage: 50,
+            genesis: ethers::types::Bytes::default(),
+        };
+        log::info!("creating subnet: {params:?}");
+
+        let call = self.registry_contract.new_subnet_actor(params);
+        let pending_tx = call.send().await?;
+        let receipt = pending_tx.retries(10).await?;
```
adlrocha, line 71: Let's add a comment explaining that we need to use retries per this issue: https://github.com/filecoin-project/community/discussions/638

src/manager/evm.rs
```diff
@@ -26,127 +30,213 @@ pub type MiddlewareImpl = SignerMiddleware<Provider<Http>, Wallet<SigningKey>>;
...
+                    match ethers_contract::parse_log::<ipc_registry::SubnetDeployedFilter>(log) {
+                        Ok(subnet_deploy) => {
+                            let ipc_registry::SubnetDeployedFilter {
+                                subnet_addr,
+                                subnet_id,
+                            } = subnet_deploy;
+
+                            log::debug!("subnet with id {subnet_id:?} deployed at {subnet_addr:?}");
+
+                            // subnet_addr.to_string() returns a summary of the actual address, not
```
adlrocha, line 86:
```suggestion
                            // subnet_addr.to_string() returns a summary of the actual Ethereum address, not
```

src/manager/evm.rs
```diff
@@ -26,127 +30,213 @@ pub type MiddlewareImpl = SignerMiddleware<Provider<Http>, Wallet<SigningKey>>;
...
     }
 
     async fn join_subnet(
         &self,
         subnet: SubnetID,
-        from: Address,
+        _from: Address,
         collateral: TokenAmount,
-        params: JoinParams,
+        _params: JoinParams,
```
adlrocha, line 111: There are no params here? Don't we need to pass the multiaddress for the validator as part of the join? If this is the case there's something wrong with the solidity contract

src/manager/evm.rs
```diff
@@ -26,127 +30,213 @@ pub type MiddlewareImpl = SignerMiddleware<Provider<Http>, Wallet<SigningKey>>;
...
+        log::info!(
+            "interacting with evm subnet contract: {address:} with collateral: {collateral:}"
+        );
+
+        let contract = SubnetContract::new(address, self.eth_client.clone());
+
+        let mut txn = contract.join();
+        txn.tx.set_value(collateral);
+
+        txn.send().await?.await?;
```
adlrocha, line 128: If the transaction doesn't go through this should fail, right?

src/manager/evm.rs
```diff
@@ -159,20 +249,124 @@ impl EthSubnetManager<MiddlewareImpl> {
...
             let wallet = priv_key.parse::<LocalWallet>()?;
             let wallet = wallet.with_chain_id(subnet.chain_id.unwrap_or_default());
 
+            let evm_gateway_address = subnet
+                .evm_gateway_address
+                .as_ref()
+                .ok_or_else(|| anyhow!("evm gateway address not defined"))?;
+            let gateway_address = H160::from_str(evm_gateway_address)?;
+
+            let evm_registry_address = subnet
```
adlrocha, line 271: Let's call this the IPC registry address throughout the code

src/manager/evm.rs
```diff
@@ -159,20 +249,124 @@ impl EthSubnetManager<MiddlewareImpl> {
...
+            return Ok(Self::new(signer, gateway_contract, evm_registry_contract));
         }
 
         Err(anyhow!(
             "no ethereum-compatible private key provided in config"
         ))
     }
 }
+
+// fn evm_id_to_address(evm_subnet: ipc_registry::SubnetID) -> Result<Address> {
```
adlrocha, line 290: Remove?

src/manager/evm.rs
```diff
@@ -159,20 +249,124 @@ impl EthSubnetManager<MiddlewareImpl> {
...
+    match payload {
+        Payload::Delegated(delegated) => {
+            let slice = delegated.subaddress();
+            Ok(ethers::types::Address::from_slice(&slice[0..20]))
+        }
+        _ => Err(anyhow!("invalid is invalid")),
+    }
+}
+
+fn agent_subnet_to_evm_addresses(subnet: &SubnetID) -> Result<Vec<ethers::types::Address>> {
```
adlrocha, line 320: Should we make it a proper From implementation? 
```suggestion
fn subnet_to_evm_address(subnet: &SubnetID) -> Result<Vec<ethers::types::Address>> {
```

src/manager/evm.rs
```diff
@@ -159,20 +249,124 @@ impl EthSubnetManager<MiddlewareImpl> {
...
+//     let mut children = vec![];
+//     let addr = evm_subnet
+//         .route
+//         .last()
+//         .ok_or_else(anyhow!("invdalid evm address passed"))?;
+//     let eth_addr = EthAddress::from_str(addr.to_string())?;
+//     Ok(Address::from(eth_addr))
+// }
+
+fn agent_subnet_to_evm_address(subnet: &SubnetID) -> Result<ethers::types::Address> {
```
adlrocha, line 301: What is the different with the function below? Can you please add comments describing what each function is?

src/manager/evm.rs
```diff
@@ -159,20 +249,124 @@ impl EthSubnetManager<MiddlewareImpl> {
...
+#[cfg(test)]
+mod tests {
+    use crate::manager::evm::{agent_subnet_to_evm_address, agent_subnet_to_evm_addresses};
+    use fvm_shared::address::Address;
+    use ipc_sdk::subnet_id::SubnetID;
+    use primitives::EthAddress;
+    use std::str::FromStr;
+
+    #[test]
+    fn test_addr_convert() {
```
adlrocha, line 337: This test has now assert, should we skip it?

src/manager/evm.rs
```diff
@@ -159,20 +249,124 @@ impl EthSubnetManager<MiddlewareImpl> {
...
+        let eth_addr = EthAddress::from_str("0x0b0d23d88d21527049232a3248fe31949d90b03b").unwrap();
+
+        let addr = Address::from(eth_addr);
+
+        println!("{addr:?}");
+    }
+
+    #[test]
+    fn test_agent_subnet_to_evm_address() {
```
adlrocha, line 346: Same here

testing/itest/src/infra/mod.rs
```diff
@@ -300,6 +300,11 @@ impl SubnetInfra {
             jsonrpc_api_ws: None,
             auth_token: Some(admin_token),
             accounts,
+            evm: false,
+            eth_private_key: None,
+            chain_id: None,
```
adlrocha, line 305: The chainID won't be needed, it can be infered from the subnetID. See: https://github.com/consensus-shipyard/ipc-actors/pull/109/files#diff-41548e2e7ed0210b54962a6cde729ac906ac66516f1c87615c4db169b3230930R71

adlrocha (changes requested):
Great job, thanks! A set of suggestions and improvements. Let's fix and clean this one, merge and test the SubnetID refactor, and try to get this one merged so we can move to the implementation of more functions with Solidity.

If you have questions we can go through the review sync.

src/server/handlers/manager/join.rs
cryptoAtwill: yeah, I would say the checking needs to be updated as well. EVM and FVM checking should be different.

Cargo.toml
cryptoAtwill: So we just use the latest master branch?

src/manager/evm.rs
cryptoAtwill: Yeah, the solidity contract does not take the validator address as parameter. Let me check with them to verify this on the latest code.

src/manager/evm.rs
cryptoAtwill: yep, it will fail.

src/manager/evm.rs
cryptoAtwill: removed

src/manager/evm.rs
cryptoAtwill: comments added

src/manager/evm.rs
cryptoAtwill: this one has assert

src/manager/evm.rs
cryptoAtwill: removed

testing/itest/src/infra/mod.rs
cryptoAtwill: updated.

cryptoAtwill:
@adlrocha I have addressed most of the comments above. BTW, probably it's a good idea to refactor `Config` as well, to separate EVM and FVM config so that we can handle the checking and parsing differently. Now they are sharing parameters and it's easy to get confused which is which, especially if we have to support both EVM and FVM simultaneously (not sure if this will happen, but just in case).

src/manager/evm.rs
```diff
@@ -26,127 +30,213 @@ pub type MiddlewareImpl = SignerMiddleware<Provider<Http>, Wallet<SigningKey>>;
...
+            }
+            None => Err(anyhow!("no receipt to event, txn not successful")),
+        };
     }
 
     async fn join_subnet(
         &self,
         subnet: SubnetID,
-        from: Address,
+        _from: Address,
```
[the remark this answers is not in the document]
cryptoAtwill: Enum is our friend.

PR #218, opened 2023-06-05
Fix itests for new subnetID and some improvements

adlrocha:
## Implementation
This PR: 
- Fixes itests to work with the new implementation of SubnetID. 
- Adds a new `log` directory where logs are persisted and `.gitignore`s unnecessary files

## Testing
In order to test the itests from this PR the branch of the following PR of eudico needs to be used: https://github.com/consensus-shipyard/lotus/pull/187

## Things to improve and high-level feedback.
@cryptoAtwill, as I was tinkering with the code I found a few things that could be improved. Would you mind having a look to see if you can address them yourself (I feel you may be more efficient as you know better this code, but let me know if you want me to do it). 
- When deploying the infra with the scripts, for some reason the resulting configuration doesn't create a new subnet (or at least doesn't appear with list subnets), and there is no way to do any tests. I don't know to what extent it is related with the address of the subnet hardcoded [here](https://github.com/consensus-shipyard/ipc-agent/blob/271f8e90722f28e7b7ac3a80753e39526f85a741/testing/itest/examples/no_create_subnet.rs#L43), what if the subnet address of the subnet created ends up being something else?
- I feel we should spend some time improve the UX of this tool (and I will also contribute to this one). The basic operation of the framework seems great, but the fact that all needs to be done manual (the deployment of the rootnet, configuring the IPC agent, etc.) is a bit of an overhead. As a next step we should include different make rules or scripts that also helps us with this (let me know if you prefer me to include it in a separate issue for future work).

### DIFF
--- a/testing/itest/src/infra/subnet.rs
+++ b/testing/itest/src/infra/subnet.rs
@@ -7,8 +7,11 @@ use anyhow::{anyhow, Result};
 use ipc_sdk::subnet_id::SubnetID;
 use std::fs;
 use std::fs::File;
+use std::path::Path;
 use std::process::{Child, Command};
 use std::thread::sleep;
+
+const DEFAULT_LOG_DIR: &str = "./logs";
 
 fn node_from_topology(topology: &SubnetConfig) -> SubnetNode {
     SubnetNode::new(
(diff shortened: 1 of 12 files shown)


testing/itest/src/infra/subnet.rs
```diff
@@ -305,10 +308,16 @@ impl SubnetNode {
 
         let subnet_id = self.subnet_id_cli_string();
 
-        let node_std_out =
-            File::create(format!("./{subnet_id:}_node_{:}.log", self.node.tcp_port))?;
-        let node_std_err =
-            File::create(format!("./{subnet_id:}_node_{:}.err", self.node.tcp_port))?;
+        let base_path = Path::new(DEFAULT_LOG_DIR);
+        fs::create_dir_all(base_path).unwrap();
```
cryptoAtwill, line 312: maybe we should use `?` here instead of `unwrap`.

cryptoAtwill (approved):

cryptoAtwill:
@adlrocha I just have one suggestion to use `?` instead of `unwrap`. 

For the not creating subnet, the reason is the file you are running is `no_create_subnet`, [here](https://github.com/consensus-shipyard/ipc-agent/blob/271f8e90722f28e7b7ac3a80753e39526f85a741/testing/itest/examples/no_create_subnet.rs#L17) it says the point is to have the subnet created prior to the execution of the script. I think your use case is more aligned with `no_tear_down` script.

testing/itest/src/infra/subnet.rs
adlrocha: Fixed

PR #219, opened 2023-06-07
fix floating point overflown

cryptoAtwill:
Fix fil parsing to u64 overflown issue so that the correct amount of fil is parsed.

### DIFF
--- a/src/server/handlers/mod.rs
+++ b/src/server/handlers/mod.rs
@@ -49,7 +49,8 @@ mod validator;
 pub mod wallet;
 
 pub type Method = String;
-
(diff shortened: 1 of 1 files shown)


src/server/handlers/mod.rs
```diff
@@ -49,7 +49,8 @@ mod validator;
 pub mod wallet;
 
 pub type Method = String;
-
+/// We only support up to 6 decimal digits for transaction
```
adlrocha:
```suggestion
/// We only support up to 9 decimal digits for transaction
```

adlrocha (approved):
LGTM.

PR #223, opened 2023-06-09
Create and join for evm

cryptoAtwill:
Support create subnet and join subnet for evm.


src/manager/evm.rs
```diff
@@ -0,0 +1,379 @@
...
+    async fn create_subnet(&self, _from: Address, params: ConstructParams) -> Result<Address> {
+        self.ensure_same_gateway(&params.ipc_gateway_addr)?;
+
+        let min_validator_stake = params
+            .min_validator_stake
+            .atto()
+            .to_u128()
+            .ok_or_else(|| anyhow!("invalid min validator stake"))?;
+
+        log::debug!("in create subnet");
```
adlrocha, line 56:
```suggestion
        log::debug!("calling create subnet for EVM manager");
```

src/manager/evm.rs
```diff
@@ -0,0 +1,379 @@
...
+
+        let mut route = agent_subnet_to_evm_addresses(&params.parent)?;
+        let mut root = vec![ethers::types::Address::from_str(
+            "0x0000000000000000000000000000000000000000",
+        )?];
+        root.append(&mut route);
+        log::debug!("root: {root:?}");
```
adlrocha, line 63:
```suggestion
        log::debug!("root SubnetID as Ethereum type: {root:?}");
```

src/manager/evm.rs
```diff
@@ -0,0 +1,379 @@
...
+        let min_validator_stake = params
+            .min_validator_stake
+            .atto()
+            .to_u128()
+            .ok_or_else(|| anyhow!("invalid min validator stake"))?;
+
+        log::debug!("in create subnet");
+
+        let mut route = agent_subnet_to_evm_addresses(&params.parent)?;
+        let mut root = vec![ethers::types::Address::from_str(
```
adlrocha, line 59: Why the first element of the route needs to be the zero address?

src/manager/evm.rs
```diff
@@ -0,0 +1,379 @@
...
+                            // subnet_addr.to_string() returns a summary of the actual Ethereum address, not
+                            // usable in the actual code.
+                            let subnet_addr = format!("{subnet_addr:?}");
+                            log::debug!("raw subnet addr: {subnet_addr:}");
+
+                            let eth_addr = EthAddress::from_str(&subnet_addr)?;
+                            return Ok(Address::from(eth_addr));
+                        }
+                        Err(_) => {
+                            log::debug!("not of event subnet actor deployed, continue");
```
adlrocha, line 109:
```suggestion
                            log::debug!("no event for subnet actor published yet, continue");
```

src/manager/evm.rs
```diff
@@ -0,0 +1,379 @@
...
+                            continue;
+                        }
+                    }
+                }
+                Err(anyhow!("no logs receipt"))
+            }
+            None => Err(anyhow!("no receipt to event, txn not successful")),
```
adlrocha, line 116:
```suggestion
            None => Err(anyhow!("no receipt for event, txn not successful")),
```

src/manager/evm.rs
```diff
@@ -0,0 +1,379 @@
...
+
+        let signer = Arc::new(SignerMiddleware::new(provider, wallet));
+        let gateway_contract = Gateway::new(gateway_address, Arc::new(signer.clone()));
+        let evm_registry_contract = SubnetRegistry::new(registry_address, Arc::new(signer.clone()));
+
+        Ok(Self::new(signer, gateway_contract, evm_registry_contract))
+    }
+}
+
+/// Convert the ipc SubnetID type to an evm address. It extracts the last address from the Subnet id
```
adlrocha, line 310: Why can't we make these two functions just one? You remove all of the childrens and then take the ones you are interested in in the relevant function? I am not a fan of functions that have a single character as the only difference in their name :)

src/manager/evm.rs
```diff
@@ -0,0 +1,379 @@
...
+
+pub struct EthSubnetManager<M: Middleware> {
+    eth_client: Arc<M>,
+    gateway_contract: Gateway<Arc<M>>,
+    registry_contract: SubnetRegistry<Arc<M>>,
+}
+
+#[async_trait]
+impl<M: Middleware + Send + Sync + 'static> SubnetManager for EthSubnetManager<M> {
+    async fn create_subnet(&self, _from: Address, params: ConstructParams) -> Result<Address> {
```
adlrocha, line 47: As discussed sync, we need to come up with some way from having to put the `Address` as part of the manager's signature. There is no way to choose a specific Address for the EVM manager, you can only use the default one.

src/server/handlers/manager/mod.rs
```diff
@@ -25,9 +26,16 @@ pub mod topdown_executed;
...
-        return Err(anyhow!("Internal server error"));
+    match &subnet.config {
+        SubnetConfig::Fvm(config) => {
+            if config.auth_token.is_none() {
+                log::error!("subnet {:?} does not have auth token", subnet.id);
+                return Err(anyhow!("Internal server error"));
+            }
+        }
+        SubnetConfig::Evm(_) => {
+            // TODO: add more checks later
```
adlrocha, line 37: What kind of checks are you thinking of?

adlrocha (approved):
LGTM overall. A few comments to address before merging, but if you have tested it and it works let's ship it :)

src/manager/evm.rs
cryptoAtwill: Yep yep, I will do the ethers wallet store then we can start using the `from`.

src/server/handlers/manager/mod.rs
cryptoAtwill: Ensure there are accounts configured (ofc after we have added the wallet store)?

src/manager/evm.rs
cryptoAtwill: Good one, I think I saw it before in the Solidity contracts, but now I think it's removed.

cryptoAtwill: updated

PR #227, opened 2023-06-14
Evm checkpoint

cryptoAtwill:
Refactoring existing checkpoint so that EVM checkpoints can be supported as well. Key changes:
- Updated `CheckpointManager` trait that removes LotusClient and all the references.
- Use `dyn CheckpointManager` in favor of specific structs for checkpoint operations.


adlrocha (approved):
LGTM. Now we just need to start with the FEVM version, but I think it makes sense to merge this one.

PR #228, opened 2023-06-15
Fevm checkpoint

cryptoAtwill:
Initial implementation of FEVM checkpoint.

Update 20230616: Add checkpoint skeleton.


src/checkpoint/fevm/bottomup.rs
```diff
@@ -0,0 +1,87 @@
...
+        &self,
+        validator: &Address,
+        epoch: ChainEpoch,
+    ) -> anyhow::Result<bool> {
+        self.parent_manager
+            .has_voted_in_subnet(&self.child_subnet.id, epoch, validator)
+            .await
+    }
+
+    async fn presubmission_check(&self) -> anyhow::Result<bool> {
```
adlrocha, line 84: What was the purpose of this function and why it is always true?

adlrocha (approved):
Overall, LGTM. But I am wondering if we can de-duplicate some code between the fevm and fvm checkpoint managers for bottom-up and top-down. I feel they both implement the same logic and what really changes is the underlying type of the subnet manager that interacts with the blockchian. WDYT?

PR #232, opened 2023-06-19
Fevm fvm checkpoint

cryptoAtwill:
Initial bottom up checkpoint submission from parent fevm to child fvm.


src/manager/evm/conversion.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+}
+
+impl TryFrom<&SubnetID> for crate::manager::evm::subnet_contract::SubnetID {
+    type Error = anyhow::Error;
+
+    fn try_from(subnet: &SubnetID) -> Result<Self, Self::Error> {
+        Ok(crate::manager::evm::subnet_contract::SubnetID {
+            root: subnet.root_id(),
+            route: agent_subnet_to_evm_addresses(subnet)?,
```
adlrocha, line 107: Can you remind me what `agent_subnet_to_evm_addresses` does? It is translating from an f4 to an 0x address? What if any of the addresses in the root is not an f4 and it can't be translated in an 0x address?

src/manager/evm/manager.rs
```diff
@@ -397,6 +397,27 @@ impl<M: Middleware + Send + Sync + 'static> EthManager for EthSubnetManager<M> {
         let initialized = self.gateway_contract.initialized().call().await?;
         Ok(initialized)
     }
+
+    async fn subnet_bottom_up_checkpoint_period(&self, subnet_id: &SubnetID) -> Result<ChainEpoch> {
+        let address = last_evm_address(subnet_id)?;
```
adlrocha, line 402: Can you remind me what we are doing with `last_evm_address`? It is gestting the subnet actor that is governing the child subnet?

adlrocha (changes requested):
A few questions and suggestions. We can discuss them sync if you think is easier. Thanks!

adlrocha (approved):

PR #233, opened 2023-06-19
Initial impl for top down

cryptoAtwill:
Add implementation for top down checkpoint submission for parent FEVM to child FVM.


src/manager/evm/topdown.rs
```diff
@@ -0,0 +1,64 @@
...
+
+use crate::manager::evm::manager::ethers_address_to_fil_address;
+use fvm_ipld_encoding::RawBytes;
+use fvm_shared::econ::TokenAmount;
+use fvm_shared::MethodNum;
+use ipc_gateway::{CrossMsg, StorableMsg};
+use ipc_sdk::address::IPCAddress;
+use ipc_sdk::subnet_id::SubnetID;
+
+impl TryFrom<crate::manager::evm::gateway::SubnetID> for SubnetID {
```
adlrocha, line 13: Isn't this type conversion already defined somewhere else? (I think I saw it in the previous PR being also implemented)

src/manager/evm/topdown.rs
```diff
@@ -0,0 +1,64 @@
...
+        let children = value
+            .route
+            .iter()
+            .map(ethers_address_to_fil_address)
+            .collect::<anyhow::Result<Vec<_>>>()?;
+        Ok(SubnetID::new(value.root, children))
+    }
+}
+
+impl TryFrom<crate::manager::evm::gateway::Ipcaddress> for IPCAddress {
```
adlrocha, line 26: Same here and for the rest of types on this file.

adlrocha (approved):
A few comments but I think it LGTM

PR #234, opened 2023-06-20
Query validator set

cryptoAtwill:
Support query validator set for EVM. Refactoring the RPC call to use common interface instead of directly calling into lotus.


adlrocha (approved):
LGTM.

PR #237, opened 2023-06-28
Integrate new subnet registry implementation

adlrocha:
As we were having trouble with emitted events in subnets, in https://github.com/consensus-shipyard/ipc-solidity-actors/pull/127 I changed the implementation of the `SubnetRegistry` so it tracked for each sender the subnets that they've deployed. With this, users can track the address of their deployed subnets, and the agent can perform the `subnet create` command without having to rely on events. 

Additional, this PR fixes minor bugs in the agent's config.


src/manager/evm/manager.rs
```diff
@@ -87,36 +87,22 @@ impl<M: Middleware + Send + Sync + 'static> SubnetManager for EthSubnetManager<M
...
-                            log::debug!("no event for subnet actor published yet, continue");
-                            continue;
-                        }
-                    }
-                }
-                Err(anyhow!("no logs receipt"))
+        call.send().await?.await?;
+
+        // getting the address
+        let sender = match self.eth_client.default_sender() {
```
cryptoAtwill, line 93: I think it might be better to use `event` as the default implementation, because the existing method only tracks the latest deployed contract, so one deploys two subnets concurrently, we might not be able to get the correct result.

PR #240, opened 2023-06-30
fix get top down messages with nonce

cryptoAtwill:
Some fixes after performing integration testing for `fund`:
- Fix get top down messages to use nonce.


src/checkpoint/fevm/topdown.rs
```diff
@@ -99,22 +99,11 @@ impl<T: EthManager + Send + Sync> CheckpointManager for TopdownCheckpointManager
...
-            .await?;
-        let checkpoint = gateway::TopDownCheckpoint {
-            epoch: epoch as u64,
-            top_down_msgs: msgs,
-        };
-        self.child_manager
-            .submit_top_down_checkpoint(checkpoint)
-            .await?;
-        Ok(())
+        todo!()
```
adlrocha, line 106: Why do we remove this one? This is the one that handles fevm-fevm, right?

src/manager/evm/manager.rs
```diff
@@ -322,7 +322,9 @@ impl<M: Middleware + Send + Sync + 'static> SubnetManager for EthSubnetManager<M
         let mut validators = vec![];
         for v in evm_validator_set.validators.into_iter() {
             validators.push(Validator {
-                addr: ethers_address_to_fil_address(&v.addr)?.to_string(),
+                // we are using worker address here so that the fvm validator node can pick up
+                // the correct
```
adlrocha, line 326:
```suggestion
                // the correct one used by the Mir validator.
```

src/checkpoint/fevm_fvm/topdown.rs
```diff
@@ -126,9 +126,17 @@ impl<P: EthManager + Send + Sync, C: LotusClient + Send + Sync> CheckpointManage
         epoch: ChainEpoch,
         validator: &Address,
     ) -> anyhow::Result<()> {
+        let nonce = gateway_state(&self.child_fvm_manager, &self.child)
```
adlrocha, line 129: Can you remind me what is the purpose of this `gateway_state` function?

adlrocha (approved):
A few comments but LGTM

src/checkpoint/fevm/topdown.rs
cryptoAtwill: Because there are quite some changes due to API update, might need some rewrite, `todo` is easier for now.

src/checkpoint/fevm_fvm/topdown.rs
cryptoAtwill: just get the state of IPC Gateway actor, this method is used over and over, put it in a function for code reuse.

PR #245, opened 2023-07-07
Add evm key store

cryptoAtwill:
Add evm key store. This implementation follows a similar approach as that in the `identity` crate. Changes:
- Add traits for key store operations
- Add memory key store
- Add persistence key store, without encryption.


eth-keystore/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+zeroize = "1.6.0"
+anyhow = { workspace = true }
+ethers = { workspace = true, optional = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+hex = "0.4.3"
```
adlrocha, line 14: Should we make this dependency part of the workspace? IIRC we are using it in other crates also.

eth-keystore/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+
+/// The key store trait for different evm key store
+pub trait KeyStore {
+    /// The type of the key that is stored
+    type Key: Clone + Eq + Hash + TryFrom<KeyInfo>;
+
+    /// Get the key info by address string
+    fn get(&self, addr: &Self::Key) -> Result<Option<KeyInfo>>;
+    /// List all addresses in the key store
+    fn list_all(&self) -> Result<Vec<Self::Key>>;
```
adlrocha, line 23: We are calling this function `list` in the Filecoin keystore, we should maybe use the same names for consistency?

eth-keystore/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+
+/// The struct that contains evm private key info
+#[derive(Clone)]
+pub struct KeyInfo {
+    private_key: Vec<u8>,
+}
+
+impl Drop for KeyInfo {
+    fn drop(&mut self) {
+        self.private_key.zeroize();
```
adlrocha, line 36: This is to avoid having the private key set in-memory, right?

eth-keystore/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+    }
+}
+
+#[cfg(feature = "with-ethers")]
```
adlrocha, line 40: Why do we need this feature?

eth-keystore/src/persistent.rs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        Ok(Self {
+            memory: MemoryKeyStore { data: key_infos },
+            file_path: path,
+        })
+    }
+
+    /// Write all keys to file without any encryption.
+    fn write_all_no_encryption(&self) -> Result<()> {
```
adlrocha, line 87: Should we call this function `flush_no_encryption`?

eth-keystore/src/persistent.rs
```diff
@@ -0,0 +1,120 @@
...
+
+        fs::create_dir_all(dir)?;
+
+        let file = File::create(&self.file_path)?;
+
+        // TODO: do we need to set path permission?
+
+        let writer = BufWriter::new(file);
+
+        let to_persist = self
```
adlrocha, line 101: It may be a bit inefficient to re-write all keys every time that we put a new identity, right? We should at least track a `TODO` here describing that this can probably be improved.

Also, we should double-check in tests that we don't write duplicates every time that we put new keys.

eth-keystore/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// The type of the key that is stored
+    type Key: Clone + Eq + Hash + TryFrom<KeyInfo>;
+
+    /// Get the key info by address string
+    fn get(&self, addr: &Self::Key) -> Result<Option<KeyInfo>>;
+    /// List all addresses in the key store
+    fn list_all(&self) -> Result<Vec<Self::Key>>;
+    /// Put a new info to the addr
+    fn put(&mut self, info: KeyInfo) -> Result<()>;
+}
```
adlrocha, line 26: We are missing a remove function. We will need it so I suggest including it already.
```suggestion
    fn remove(&mut self, addr: &Self::Key)
}
```

adlrocha (changes requested):
Overall LGTM,  but we should introduce a `remove` function. I also left a few other comments/suggestions. We can discuss sync if needed.

eth-keystore/src/lib.rs
cryptoAtwill: The key to the key store does not assume the type of the key, so we dont have a hard dependency on `ethers`. So that we can expand to other crates, just more options and dont force downstream to use ethers.

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = [".", "testing/e2e", "identity", "testing/itest"]
+members = [".", "testing/e2e", "identity", "testing/itest", "eth-keystore"]
```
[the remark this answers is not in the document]
cryptoAtwill: yeah, initially I was thinking along this line, but when I started to implement it, I realised it would make the `identity` crate quite messy or need some refactoring, like renaming some struct and move around the files. So I started with a new crate.

eth-keystore/src/lib.rs
adlrocha: I like it!

Cargo.toml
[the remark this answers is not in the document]
adlrocha: I see... I think it is worth doing the effort now of following that approach, it may be harder to refactor in the future, WDYT?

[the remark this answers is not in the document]
cryptoAtwill: this works, time to see how good the IDE is.

eth-keystore/src/persistent.rs
cryptoAtwill: I followed the implementation of `identity`, copied and adapted the tests over.

adlrocha (approved):

PR #246, opened 2023-07-07
docs to deploy IPC Solidity in calibration

adlrocha:
This PR adds a quickstart guide to deploy IPC solidity contracts to Calibration. Additionally, it updates the config and cross-net message primitives with all the changes to be included in the next release.

### DIFF
--- a/docs/calibration-solidity.md
+++ b/docs/calibration-solidity.md
@@ -21,6 +21,8 @@ At a high-level, the steps that you need to follow are the following (for those 
 
 With this, we have the funds we need to start deploying IPC.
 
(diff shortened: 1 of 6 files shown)


docs/calibration-solidity.md
```diff
@@ -0,0 +1,229 @@
...
+```toml
+[server]
+json_rpc_address = "0.0.0.0:3030"
+
+[[subnets]]
+id = "/r314159"
+network_name = "calibration"
+
+[subnets.config]
+accounts = ["0x6BE1Ccf648c74800380d0520D797a170c808b624"]
```
cryptoAtwill, line 85: Since we are integrating evm key stpre now, accounts will auto use the address of the private key, so even if it is configured, I think it will not take effect. But once the PR for key store is merged, `account` will behave exactly like that in fvm and `private_key` will be deprecated.

adlrocha: Let's update the docs as part of that PR if you don't mind. Should we make that PR to target this one?

cryptoAtwill: The evm key store PR is in progess, once done we can just update the doc or even in a new PR with only the doc updates, should be easier to review that way.

jsoares (approved):
Looke fine. Still need tp properly testnet nit looks good to merge.

docs/calibration-solidity.md
adlrocha: Fixed

PR #249, opened 2023-07-10
Integrate evm key store

cryptoAtwill:
Integrate evm key store into the existing code base. Changes:
- Evm Subnet manager dynamically create contract instance upon receiving sender parameter.
- Removed Evm subnet manager `Middleware` to use `SignerMiddleware` directly because we need to use `SignerMiddleware` directly.
- Update the impacted constructor due to the previous change.


src/checkpoint/fevm/bottomup.rs
```diff
@@ -113,7 +113,7 @@ impl<T: EthManager + Send + Sync, M: LotusClient + Send + Sync> CheckpointManage
         &self,
         epoch: ChainEpoch,
         // TODO: when we support more wallet addresses, we need this variable
```
adlrocha, line 115: Remove TODO? 
```suggestion
```

src/checkpoint/mod.rs
```diff
@@ -76,15 +77,21 @@ pub struct CheckpointSubsystem {
     /// The subsystem uses a `ReloadableConfig` to ensure that, at all, times, the subnets under
     /// management are those in the latest version of the config.
     config: Arc<ReloadableConfig>,
-    wallet_store: Arc<RwLock<Wallet>>,
+    fvm_keystore: Arc<RwLock<Wallet>>,
```
adlrocha, line 80: We should figure out a better naming and add consistency to the implementation of both keystores. I feel they should have a common interface and similar namings.

src/manager/evm/manager.rs
```diff
@@ -33,21 +34,29 @@ pub type MiddlewareImpl = SignerMiddleware<Provider<Http>, Wallet<SigningKey>>;
...
-    eth_client: Arc<M>,
-    gateway_contract: Gateway<Arc<M>>,
-    registry_contract: SubnetRegistry<Arc<M>>,
+pub struct EthSubnetManager {
+    keystore: Arc<RwLock<PersistentKeyStore<ethers::types::Address>>>,
+    on_chain_info: OnChainInfo,
+}
+
+/// Keep track of the on chain information for the subnet manager
+struct OnChainInfo {
```
adlrocha, line 50: What about calling it `IPCInfo` or `IPCConfig`? 
```suggestion
struct IPCInfo {
```

src/manager/evm/manager.rs
```diff
@@ -533,36 +560,97 @@ impl<M: Middleware + Send + Sync + 'static> EthManager for EthSubnetManager<M> {
...
+    fn get_ethers_client(&self, addr: Option<&ethers::types::Address>) -> Result<MiddlewareImpl> {
+        let wallet = if let Some(addr) = addr {
+            let keystore = self.keystore.read().unwrap();
+            let private_key = keystore
+                .get(addr)?
+                .ok_or_else(|| anyhow!("address {addr:} does not have private key in key store"))?;
+            LocalWallet::from_bytes(private_key.private_key())?
+                .with_chain_id(self.on_chain_info.chain_id)
+        } else {
+            LocalWallet::from_bytes(&RANDOM_PRIVATE_KEY)?
```
adlrocha, line 639: Is this the right way to do it? Can't we initialize here a `Middleware` instead of a `SignerMiddleware` to call the `view` functions? I feel generating a new random key every time can be really inefficient.

adlrocha (changes requested):
Nice work. A few suggestons and comments. Before we merge, something we should also do is remove the `private_key` field from the config if it is no longer used (from the template, type definitions, and tests)

src/manager/evm/manager.rs
```diff
@@ -27,7 +28,8 @@ use crate::lotus::message::ipc::{QueryValidatorSetResponse, SubnetInfo, Validato
 use crate::lotus::message::wallet::WalletKeyType;
 use crate::manager::{EthManager, SubnetManager};
 
-pub type MiddlewareImpl = SignerMiddleware<Provider<Http>, Wallet<SigningKey>>;
+pub type NonSignerM = Provider<Http>;
```
adlrocha, line 31: Are these used? If so let's use maybe `RawMiddleware` and `SignerMiddleware` so they are self-describing.

adlrocha (approved):
LGTM, I think all the comments are addressed. If you tested it and it works let's merge.

src/manager/evm/manager.rs
cryptoAtwill: This is updated.

src/manager/evm/manager.rs
cryptoAtwill: This issue no longer present in latest code.

src/manager/evm/manager.rs
cryptoAtwill: We have two different middleware now, so it's easier for typing, I will update the naming accordingly, maybe we can have a better name than `SignerMiddleware` because it's taken by the `ethers` crate.

PR #251, opened 2023-07-11
Import evm wallet

cryptoAtwill:
Add import evm wallet to RPC and cli functions.

### DIFF
--- a/identity/src/evm/persistent.rs
+++ b/identity/src/evm/persistent.rs
@@ -51,7 +51,6 @@ impl<T: Clone + Eq + Hash + AsRef<[u8]> + TryFrom<KeyInfo>> KeyStore for Persist
 
     fn put(&mut self, info: KeyInfo) -> Result<Self::Key> {
         let addr = self.memory.put(info)?;
-        // TODO: We can flush to disk only after certain number of `put` is called.
         self.flush_no_encryption()?;
         Ok(addr)
     }
--- a/identity/src/evm/persistent.rs
+++ b/identity/src/evm/persistent.rs
@@ -49,10 +49,11 @@ impl<T: Clone + Eq + Hash + AsRef<[u8]> + TryFrom<KeyInfo>> KeyStore for Persist
         self.memory.list()
     }
 
-    fn put(&mut self, info: KeyInfo) -> Result<()> {
-        self.memory.put(info)?;
+    fn put(&mut self, info: KeyInfo) -> Result<Self::Key> {
+        let addr = self.memory.put(info)?;
         // TODO: We can flush to disk only after certain number of `put` is called.
-        self.flush_no_encryption()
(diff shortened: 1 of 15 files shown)


identity/src/evm/persistent.rs
```diff
@@ -49,10 +49,11 @@ impl<T: Clone + Eq + Hash + AsRef<[u8]> + TryFrom<KeyInfo>> KeyStore for Persist
         self.memory.list()
     }
 
-    fn put(&mut self, info: KeyInfo) -> Result<()> {
-        self.memory.put(info)?;
+    fn put(&mut self, info: KeyInfo) -> Result<Self::Key> {
+        let addr = self.memory.put(info)?;
         // TODO: We can flush to disk only after certain number of `put` is called.
```
adlrocha, line 54: Would you mind explaining me what do you mean with this TODO?

src/cli/commands/wallet/import.rs
```diff
@@ -11,6 +12,9 @@ use crate::cli::commands::get_ipc_agent_url;
 use crate::cli::{CommandLineHandler, GlobalArguments};
 use crate::sdk::{IpcAgentClient, LotusJsonKeyType};
 
+const FVM_WALLET: u8 = 0;
```
adlrocha, line 15: Why not an enum?

src/cli/commands/wallet/import.rs
```diff
@@ -45,6 +55,12 @@ impl CommandLineHandler for WalletImport {
 pub(crate) struct WalletImportArgs {
     #[arg(long, short, help = "The JSON RPC server url for ipc agent")]
     pub ipc_agent_url: Option<String>,
+    #[arg(
+        long,
+        short,
+        help = "The network the key belongs to, 0 for fvm, 1 for evm"
+    )]
+    pub network_type: u8,
```
adlrocha, line 63: Would it be possible to make it an enum or even two different boolean flags `--fvm` and `--evm`?

adlrocha (approved):
LGTM. I added a few suggestions, and one question, should we implement also the import for fvm and evm? 

Also, did you test the CLI or only through the RPC server? Thanks!

src/cli/commands/wallet/import.rs
cryptoAtwill: I think we might not be able to pass enum to clap/cli, or at least not that easy. So use a number is just easier.

identity/src/evm/persistent.rs
cryptoAtwill: I mean to flush to disk only after some buffer mechanism, so that we dont have to flush every time when we call put.

adlrocha, line 54:
```suggestion
```

PR #256, opened 2023-07-13
Refactor bottom up

cryptoAtwill:
Initial refactoring of bottom up, some changes:
- Introduced `NativeBottomUpCheckpoint` struct that can be converted to all different network types.
- Introduced `BottomUpManager` struct that take `BottomUpHandler` to handle checkpoints
- Let `LotusSubnetManager` and `EvmSubnetManager` implement `BottomUpHandler`.


src/manager/lotus.rs
```diff
@@ -434,11 +461,233 @@ impl<T: JsonRpcClient + Send + Sync> LotusSubnetManager<T> {
...
+                checkpoint.source
+            )
+        })?;
+        log::debug!("checkpoint message published with cid: {message_cid:?}");
+
+        Ok(self.lotus_client.state_wait_msg(message_cid).await?.height as ChainEpoch)
+    }
+}
+
+mod conversion {
```
adlrocha, line 614: Should we take this module into `fvm/conversion.rs` like we did with `evm/conversion.rs` or maybe add a `conversion` subfolder that has `conversion/fvm.rs` and `conversion/evm.rs`?

src/checkpoint/mod.rs
```diff
@@ -16,13 +16,16 @@ use tokio::select;
...
 
-pub use fvm::*;
+pub use bottomup::*;
 use ipc_identity::PersistentKeyStore;
+use ipc_sdk::subnet_id::SubnetID;
+pub use proof::create_proof;
 use std::fmt::Display;
 
+mod bottomup;
 mod fevm;
```
adlrocha, line 26: Can't we remove all of these modules and maybe just have `evm` and `fvm` and choose the right checkpoint manager implementation according to the config? (I think this is where we are going once we refactor top-down, but just to double-check).

adlrocha (approved):
A few comments and suggested changes. Just as a sanity-check, this is only organizing the code and moving function implementations around, right? I haven't dig deep into the function implementations, but let me know if I should do so. Thanks.

src/checkpoint/mod.rs
```diff
@@ -32,6 +35,18 @@ const SUBMISSION_LOOK_AHEAD_EPOCH: ChainEpoch = 50;
 /// Checkpoint manager that handles a specific parent - child - checkpoint type tuple.
 /// For example, we might have `/r123` subnet and `/r123/t01` as child, one implementation of manager
 /// is handling the top-down checkpoint submission for `/r123` to `/r123/t01`.
+#[async_trait]
+pub trait VoteQuery<T> {
```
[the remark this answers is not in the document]
cryptoAtwill: yep, TopDown will have something like this as well

src/checkpoint/setup.rs
```diff
@@ -74,14 +75,18 @@ async fn parent_fevm_child_fvm(
...
+    let fevm = FevmSubnetManager::new(
+        EthSubnetManager::from_subnet_with_wallet_store(parent, evm_wallet_store.clone())?,
+        LotusJsonRPCClient::from_subnet_with_wallet_store(child, fvm_wallet_store.clone()),
+    );
+    let fvm = LotusSubnetManager::new(
+        LotusJsonRPCClient::from_subnet_with_wallet_store(child, fvm_wallet_store.clone()),
+        child.gateway_addr(),
+    );
     let m: Box<dyn CheckpointManager> = Box::new(
-        crate::checkpoint::fevm_fvm::BottomUpCheckpointManager::new(
```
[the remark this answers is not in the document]
cryptoAtwill: yep, once I implemented top down, will remove all the combinations

src/checkpoint/mod.rs
cryptoAtwill: yeah, that should be possible

src/manager/lotus.rs
cryptoAtwill: yep, this works too. Initially I thought it's just a small mod, so just placed in place. Separating into another file works fine.

src/checkpoint/bottomup.rs
```diff
@@ -0,0 +1,181 @@
...
+use crate::config::Subnet;
+use anyhow::{anyhow, Result};
+use async_trait::async_trait;
+use fvm_shared::address::Address;
+use fvm_shared::clock::ChainEpoch;
+use ipc_gateway::checkpoint::BatchCrossMsgs;
+use ipc_sdk::subnet_id::SubnetID;
+use std::fmt::{Display, Formatter};
+
+pub type Bytes = Vec<u8>;
```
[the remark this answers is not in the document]
cryptoAtwill: updated

src/checkpoint/mod.rs
adlrocha: We should defer it for after the top-down checkpoint is in?

PR #259, opened 2023-07-18
add support for wallet remove and fevm balance list

adlrocha:
This PR: 
- Adds a `wallet remove` command to allow the removal of identities from the wallet. 
- It adds support for `wallet balance` for FEVM-based subnets.
- It refactors and improves the code in a few places here and there.

### DIFF
--- a/src/manager/evm/manager.rs
+++ b/src/manager/evm/manager.rs
@@ -1,9 +1,9 @@
 // Copyright 2022-2023 Protocol Labs
 // SPDX-License-Identifier: MIT
 use std::collections::HashMap;
-use std::str::FromStr;
 use std::sync::{Arc, RwLock};
 
+pub use crate::manager::evm::{ethers_address_to_fil_address, fil_to_eth_amount};
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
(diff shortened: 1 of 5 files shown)


cryptoAtwill (approved):

src/manager/evm/manager.rs
```diff
@@ -302,16 +298,14 @@ impl SubnetManager for EthSubnetManager {
...
-    async fn wallet_balance(&self, _address: &Address) -> Result<TokenAmount> {
-        todo!()
+    async fn wallet_balance(&self, address: &Address) -> Result<TokenAmount> {
+        let balance = self
+            .ipc_contract_info
+            .provider
+            .clone()
+            .get_balance(payload_to_evm_address(address.payload())?, None)
+            .await?;
+        Ok(TokenAmount::from_atto(balance.as_u128()))
```
cryptoAtwill, line 308: in the `send_value` PR, we have a `eth_to_fil_amount` method that handles the conversion as str. It should be better than using `u128` as intermediate type.

adlrocha: I see you already fixed it there :) Thanks!

PR #266, opened 2023-07-26
Integrate diamond

cryptoAtwill:
Integrate with solidity diamond pattern, some key changes:
- Removed `gateway` and `subnet actor` abi and replaced with diamond abi
- Updated `conversion` module that uses macro rules to generate types so that code duplication is minimized.
This PR relies on https://github.com/consensus-shipyard/ipc-solidity-actors/pull/158 to provide full getter functions.


adlrocha (approved):
Nice!! LGTM.

PR #267, opened 2023-07-27
Implement list subnet

cryptoAtwill:
Implement list subnet for solidity contracts. Key changes:
- Removed `serialization` module and migrated code to `NativeBottomUpCheckpoint`
- Added corresponding contract call
- Updated interface to use `NativeBottomUpCheckpoint` and updated different type conversions.
- Introduced `bottom_up_checkpoint` macro rules to handle different solidity contract generated module type conversion


src/checkpoint/bottomup.rs
```diff
@@ -180,3 +184,132 @@ impl<P: BottomUpHandler, C: BottomUpHandler> CheckpointManager for BottomUpManag
         Ok(true)
     }
 }
+
+// Serialization related
```
adlrocha, line 188: Out of a curiosity, why are these needed?

adlrocha (approved):
LGTM!

PR #269, opened 2023-07-27
Add action to check and fix PR targets

jsoares:
@adlrocha following slack discussion: this would allow us to retain main as the default while automatically redirecting in case some rattlebrain like me targets the wrong branch.


adlrocha (approved):
Good one! Thanks! I will revert the change in the main branch

PR #270, opened 2023-07-28
Fix quickstart wallet configuration

jsoares:
Fixes incomplete instructions leading to reported issues with empty set of validators for bottom-up checkpoints:
- Drops primary wallet creation step, since this wasn't used for anything
- Adds 3 owner wallets to agent config

Targets `main` given current state of affairs and @adlrocha's intent to merge.


adlrocha (approved):
Thank you for fixing this one. Hopefully once we deprecate native contracts it won't be as confusing :crossed_fingers:

PR #279, opened 2023-08-07
Integrate messager facet

cryptoAtwill:
Integrate messager facet from the latest solidity contract. Also removing the deprecated whitelist propagator.


src/cli/commands/crossmsg/mod.rs
```diff
@@ -31,7 +31,7 @@ impl CrossMsgsCommandsArgs {
             Commands::Fund(args) => Fund::handle(global, args).await,
             Commands::Release(args) => Release::handle(global, args).await,
             Commands::Propagate(args) => Propagate::handle(global, args).await,
-            Commands::WhitelistPropagator(args) => WhitelistPropagator::handle(global, args).await,
+            Commands::SendCrossMsgs(args) => SendCrossMsg::handle(global, args).await,
```
adlrocha, line 34: Hmm.. `SendCross` was supposed to be an internal function from the smart contract not accessible by externally owned accounts, but now that I see it here I think it is a good idea to also expose it in the agent. It may come handy for some users.

src/cli/commands/crossmsg/send_cross.rs
```diff
@@ -47,15 +46,13 @@ impl CommandLineHandler for WhitelistPropagator {
...
     pub ipc_agent_url: Option<String>,
     #[arg(long, short, help = "The address that owns the message in the subnet")]
     pub from: Option<String>,
     #[arg(long, short, help = "The subnet to whitelist")]
     pub subnet: String,
-    #[arg(help = "The message cid to whitelist")]
-    pub postbox_msg_cid: Cid,
-    #[arg(help = "The addresses to whitelist")]
-    pub to_add: Vec<String>,
+    #[arg(help = "The cross network message to send")]
```
adlrocha, line 56: This is wrong, in `SendCross`  the argument is a `Message` that one wants to send to a subnet, not a string. As discussed, let's maybe remove this code for now, or add a comment with a huge `FIXME` at the beginning of the file, but not expose this to users.

src/manager/fvm/mod.rs
```diff
@@ -227,18 +226,14 @@ impl<T: JsonRpcClient + Send + Sync> SubnetManager for LotusSubnetManager<T> {
         subnet: SubnetID,
         gateway_addr: Address,
         from: Address,
-        postbox_msg_cid: Cid,
+        postbox_msg_cid: Vec<u8>,
```
adlrocha, line 229: Shouldn't we keep it as a `cid` and make it into a `Vec<u8>` for the solidity contract? I think it is confusing to pass here a `Vec<u8>` when it actually is a `Cid` under the hood.

adlrocha (approved):
A few things before we merge:
- I wouldn't expose publicly in the CLI `sendCross` to users as it may be quite confusing. Initially, we are thinking as `sendCross` to be a function only called by smart contracts in a subnet. This may change in the future, but as we haven't tested it yet, I would keep it "out of sight" for now :)
- I think it makes more sense for the argument of `propagate` to be a `cid`  instead of `Vec<u8>` if under the hood we are using a `cid`. I see that in the contract implementation we expect a bytes, maybe we should double check if this is a hash or a cid and do accordingly (the only issue will arise in the interoperability between native and Solidity contracts, as always). 

Apart from this, if the e2e lifecycle of subnets and fund/release work we can get this merged.

src/cli/commands/crossmsg/mod.rs
cryptoAtwill: maybe just RPC calls then

src/manager/fvm/mod.rs
cryptoAtwill: In solidity, it's `bytes32` and it's derived from `keccak`. As discussed, we will rename the parameter to `postbox_msg_key` and use `Vec<u8>`.

src/manager/subnet.rs
```diff
@@ -71,31 +71,29 @@ pub trait SubnetManager: Send + Sync {
...
-    /// Propagate a cross-net message forward
+    /// Propagate a cross-net message forward. For `postbox_msg_cid`, we are using bytes because different
+    /// runtime have different representations.
     async fn propagate(
         &self,
         subnet: SubnetID,
         gateway_addr: Address,
         from: Address,
-        postbox_msg_cid: Cid,
+        postbox_msg_cid: Vec<u8>,
```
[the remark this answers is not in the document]
cryptoAtwill: same as the comment above

src/manager/evm/manager.rs
```diff
@@ -297,29 +300,60 @@ impl SubnetManager for EthSubnetManager {
...
+        gateway_contract
+            .send_cross_message(evm_cross_msg)
+            .send()
+            .await?;
+
+        Ok(())
     }
 
-    async fn whitelist_propagator(
+    async fn set_validator_net_addr(
```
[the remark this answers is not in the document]
cryptoAtwill: This method is not implemented yet in the contract.

adlrocha (approved):

PR #283, opened 2023-08-08
Fix checkpoint look ahead

### LINKED ISSUE #284: Hardcoded look ahead limit prevents checkpoint submission with large period
### Issue type

Bug

Yes

### Version

v0.5.0

### Custom code

Yes

Linux Ubuntu 22.04

### Describe the issue

https://github.com/consensus-shipyard/ipc-agent/blob/88d76bb618540619ddf3a15a83b6d1f546587728/src/checkpoint/mod.rs#L250-L261

We use a hardcoded look ahead limit to rate-limit checkpoint generation. If the checkpointing period is larger than the look-ahead period _or_ if the checkpointing process falls behind and crosses the threshold, we enter a stable equilibrium where the `while` condition will always return false and no new checkpoints will be generated.

```rust
while next_epoch < cut_off_epoch {
   ...
}
```
will be expanded to:
```rust
while (last_executed_epoch + period) < (SUBMISSION_LOOK_AHEAD_EPOCH + last_executed_epoch) {
    ...
}
```
which is:
```
while period < SUBMISSION_LOOK_AHEAD_EPOCH {
    ...
}

### Repro steps

1. Create a subnet with bottom-up checkpointing period 900
2. Run the validators and agent

### Relevant log output

```shell
[2023-08-07T23:32:17Z INFO  ipc_agent::checkpoint] latest epoch 2652, last executed epoch: 0 for checkpointing: bottom-up, parent: /r314159, child: /r314159/t410fqjlaperrmrkxna5xo7mpzfkl7l3ofoob63s5npy
[2023-08-07T23:32:17Z INFO  ipc_agent::checkpoint] process checkpoint from epoch: 0 to 2652 in manager: bottom-up, parent: /r314159, child: /r314159/t410fqjlaperrmrkxna5xo7mpzfkl7l3ofoob63s5npy
```

### Fix 

* An obvious mitigation is to just increase the look-ahead limit. We definitely need it to be higher: it's become clear that we're going to have to support much higher checkpointing frequencies, but that's not enough. Any solution that relies on a naive comparison to a max epoch delta will enter a bad stable equilibrium if the threshold is crossed (because the agent was down for some time, or whatever other reason).
* It is more intuitive and robust to specify the look-ahead period in number of checkpoints rather than in number of epochs, i.e. cap the checkpoint batch size per function call.

cryptoAtwill:
# Motivation

Fixes #284 

Previously checkpoint look ahead is using a fix epoch number, the issue is when the checkpoint period is larger than the checkpoint look ahead, the checkpoint submission logic is not triggered.
```rust
while next_epoch < cut_off_epoch {
   ...
}
```
will be expanded to:
```rust
while (last_executed_epoch + period) < (SUBMISSION_LOOK_AHEAD_EPOCH + last_executed_epoch) {
    ...
}
```
which is:
```
while period < SUBMISSION_LOOK_AHEAD_EPOCH {
    ...
}
```

# Fix
Fix checkpoint look ahead using multiple of checkpoint period will resolve the issue. 
```rust
let cut_off_epoch = std::cmp::min(
    current_epoch,
    SUBMISSION_LOOK_AHEAD_ROUNDS * period + last_executed_epoch,
);
```

Current `SUBMISSION_LOOK_AHEAD_ROUNDS` is set to 5, we can adjust this number in the future as we see more fit.

### DIFF
--- a/src/checkpoint/mod.rs
+++ b/src/checkpoint/mod.rs
@@ -29,7 +29,7 @@ mod setup;
 mod topdown;
 
 const TASKS_PROCESS_THRESHOLD_SEC: u64 = 15;
-const SUBMISSION_LOOK_AHEAD_EPOCH: ChainEpoch = 50;
+const SUBMISSION_LOOK_AHEAD_ROUNDS: i64 = 3;
 
 /// Checkpoint manager that handles a specific parent - child - checkpoint type tuple.
 /// For example, we might have `/r123` subnet and `/r123/t01` as child, one implementation of manager
@@ -250,7 +250,7 @@ async fn submit_till_current_epoch(manager: &dyn CheckpointManager) -> Result<()
     let mut next_epoch = last_executed_epoch + period;
     let cut_off_epoch = std::cmp::min(
         current_epoch,
-        SUBMISSION_LOOK_AHEAD_EPOCH + last_executed_epoch,
+        SUBMISSION_LOOK_AHEAD_ROUNDS * period + last_executed_epoch,
     );
 
     // Instead of loop all the way to `current_epoch`, we loop till `cut_off_epoch`.


adlrocha (approved):
I think this fix makes sense. We still may have a few issues:
- Like the epoch being too long in the past and facing the lookback limitation of some gateways.
- With lots of validators managed by the agent (which shouldn't be the case, but who knows), the fact that the catch-up is done sequentially may lead to not being able to ever fully catch-up (this is not a huge issue, but it could lead to additional delays in cross-net messages) 

That being said, with the new cross-net primitives we are considering for Fendermint the aforementioned issues will be removed, so I don't think it is addressing them right now. 

Sorry for the rant, TL;DR: LGTM so feel free to merge :)

jsoares:
Have we confirmed that the fix works and is complete? It seems like we may have bumped into a different bug when trying to catch up the `alpha` subnet using this fix, but I'd like to be sure that the two are unrelated.

adlrocha:
> It seems like we may have bumped into a different bug when trying to catch up the alpha subnet using this fix, but I'd like to be sure that the two are unrelated.

Oh, I think missed this one. @cryptoAtwill would you mind briefing us on what is the status? Thanks!

jsoares:
I created a new actors issue to track the signing problem: https://github.com/consensus-shipyard/ipc-solidity-actors/issues/188

If we're convinced the two things are unrelated, we can go ahead with merging this PR.

jsoares (approved):

cryptoAtwill:
@jsoares I tried again locally, I see contract reverted error. I think this PR is not related to #188, we can merge this one first.

cryptoAtwill:
@jsoares I tried again locally, I see contract reverted error. I think this PR is not related to https://github.com/consensus-shipyard/ipc-agent/pull/188, we can merge this one first.

PR #287, opened 2023-08-09
IPC-287: Fix high premium in eth_call

adlrocha:
__Note: Do not merge yet, I want to thoroughly test before merging__

Fixes https://github.com/consensus-shipyard/ipc-agent/issues/276

`ethers` performs several Ethereum-specific assumptions to estimate the gas premium. When the premium is too low, it sets a default premium that corresponds to `3 nanoFiL`. The average premiums paid in Filecoin are lower than the ones in Filecoin, what led to `ethers` always setting the premium to the default value (quite high for the case of Filecoin). 

This PR addresses the issue by using a custom premium estimation based on the historical premium paid in the last few blocks. Unlike `ethers` we don't use a threshold trigger that sets the default premium. This gas estimation is used for every  `eth_call` from the agent.

### DIFF
--- a/src/manager/evm/manager.rs
+++ b/src/manager/evm/manager.rs
@@ -13,7 +13,10 @@ use ethers::prelude::k256::ecdsa::SigningKey;
 use ethers::prelude::{abigen, Signer, SignerMiddleware};
 use ethers::providers::{Authorization, Http, Middleware, Provider};
 use ethers::signers::{LocalWallet, Wallet};
-use ethers::types::Eip1559TransactionRequest;
+use ethers::types::{Eip1559TransactionRequest, I256, U256};
+use ethers::utils::{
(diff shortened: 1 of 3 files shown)


src/manager/evm/manager.rs
```diff
@@ -834,6 +856,124 @@ impl EthSubnetManager {
...
+    if base_fee_per_gas <= U256::from(40_000_000_000u64) {
+        base_fee_per_gas * 2
+    } else if base_fee_per_gas <= U256::from(100_000_000_000u64) {
+        base_fee_per_gas * 16 / 10
+    } else if base_fee_per_gas <= U256::from(200_000_000_000u64) {
+        base_fee_per_gas * 14 / 10
+    } else {
+        base_fee_per_gas * 12 / 10
+    }
+}
```
jsoares, line 920: This is some quality logic 😓 (I realise it's borrowed)

Any concerns re how these constants translate to Filecoin vs eth? It's probably fine.

jsoares (approved):
Looks reasonable enough. If only my bottom-up checkpoints worked right now, I'd even validate it :)

cryptoAtwill (approved):

PR #288, opened 2023-08-09
Support setValidatorNetAddr, setValidatorWorkerAddr, listValidators

dnkolegov:
This PR fixes https://github.com/consensus-shipyard/ipc-agent/issues/286  and https://github.com/consensus-shipyard/ipc-agent/issues/291 by adding support for `setValidatorNetAddr`,  `setValidatorWorkerAddr`, and `listValidators`.


src/manager/evm/manager.rs
```diff
@@ -353,11 +353,43 @@ impl SubnetManager for EthSubnetManager {
...
+    ) -> Result<()> {
+        let address = contract_address_from_subnet(&subnet)?;
+        log::info!(
+            "set validator net addr: {net_addr:} on evm subnet: {subnet:} at contract: {address:}"
+        );
+
+        let signer = self.get_signer(&from)?;
+        let contract = SubnetActorManagerFacet::new(address, Arc::new(signer));
+
+        let txn = contract.set_validator_net_addr(net_addr);
```
adlrocha, line 368: Have a look at the latest PR I merged. We need to make the call with the premium estimation to avoid over-paying on the inclusion of the message: https://github.com/consensus-shipyard/ipc-agent/pull/287/files#diff-bea17d53f080353906f6ed16be5c1ed1d4365bad853a9c4db92e8287d59e654fR216

src/manager/evm/manager.rs
```diff
@@ -353,11 +353,43 @@ impl SubnetManager for EthSubnetManager {
...
+        worker_addr: Address,
     ) -> Result<()> {
-        todo!()
+        let address = contract_address_from_subnet(&subnet)?;
+        log::info!("set validator worker addr: {worker_addr:} on evm subnet: {subnet:} at contract: {address:}");
+
+        let signer = self.get_signer(&from)?;
+        let contract = SubnetActorManagerFacet::new(address, Arc::new(signer));
+
+        let txn = contract
```
adlrocha, line 387: Same as above, add the premium estimation here.

adlrocha (changes requested):
This is great, thank you Denis. I need you to update the contract calls to include the premium estimation as in the recent PR I merged to avoid over-paying in these transactions, but apart from that looks good.

Would you mind testing that it works before merging it? Thanks!

src/manager/evm/manager.rs
dnkolegov: Fixed

src/manager/evm/manager.rs
dnkolegov: Fixed

cryptoAtwill (approved):

adlrocha (approved):

PR #294, opened 2023-08-12
Add gh action to upload binaries on release

### LINKED ISSUE #292: Generate binaries for agent releases
### Issue type

Build/Install

No

### Version

n/a

### Custom code

No

n/a

### Describe the issue

We currently only release source code and require users to build the agent. For a user not deploying their own subnet, distributing binaries would greatly reduce the time to get started with IPC.

### Repro steps

n/a

### Relevant log output

_No response_

jsoares:
On release, builds and attaches binaries for linux x64 and darwin universal (arm+x64).

Also replaces deprecated action (see https://github.com/consensus-shipyard/fendermint/pull/188#discussion_r1295111695)

Closes #292

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,9 +17,8 @@ jobs:
       - name: Check out the project
         uses: actions/checkout@v3
       - name: Install Rust
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@master
         with:
-          profile: minimal
           toolchain: nightly
           components: rustfmt
       - name: Check code formatting
@@ -62,9 +61,8 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install Rust
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@master
         with:
-          profile: minimal
           target: wasm32-unknown-unknown
           toolchain: ${{ matrix.rust }}
           components: rustfmt,clippy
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -62,9 +61,8 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install Rust
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@master
         with:
-          profile: minimal
           target: wasm32-unknown-unknown
```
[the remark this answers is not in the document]
jsoares: oops, thx!

.github/workflows/release-binaries.yaml
```diff
@@ -0,0 +1,29 @@
+name: Release binaries
+
+permissions:
+  contents: write
+
+on:
+  workflow_dispatch:
+  release:
+    types: [ created ]
```
aakoshh, line 9: Would it be worth also uploading binaries on tagging? I'm not familiar with the release process. 

```
on:
  push:
    tags:
      - v[0-9]+.*
```

aakoshh (approved):
After briefly looking at action docs, this looks good to me :+1:

.github/workflows/release-binaries.yaml
aakoshh: Never mind, I see the release can create a tag and it's just better because you have to describe what you did.

PR #295, opened 2023-08-22
Update sdk for fendermint

cryptoAtwill:
Update the `ipc-sdk` for fendermint top down checkpoint. 

Key changes:
- Support `get_chain_head`, `get_top_down_msgs`, `get_block_hash` rpc calls.
- Update `queryValidatorSet` with extra optional parameter: epoch. Idea is to query the validator set up till that epoch.


ipc-agent-sdk/Cargo.toml
adlrocha, line 1: What about calling it `ipc-sdk` and move here all the types that are common from the `ipc-actors`? At least until we decide if we should start a new independent repo for the `ipc-sdk`.

ipc-agent-sdk/src/apis/topdown.rs
```diff
@@ -0,0 +1,73 @@
...
+                serde_json::json!({
+                    "subnet_id": subnet_id.to_string(),
+                    "start": start,
+                    "to": to,
+                }),
+            )
+            .await
+    }
+
+    pub async fn get_chain_head(&self, subnet_id: String) -> anyhow::Result<ChainHeadResponse> {
```
adlrocha, line 31: The other day I was chatting with Jorge, and for those functions that directly interact with the subnet without requiring anything from the agent we should consider moving the RPC implementation to the SDK and make the all directly. That way we can use the `ipc-agent` without requiring a `daemon` running. I think this could really improve the UX.

ipc-agent-sdk/src/message/chain.rs
```diff
@@ -0,0 +1,16 @@
+// Copyright 2022-2023 Protocol Labs
+// SPDX-License-Identifier: MIT
+//! Contains the chain related RPC request and response messages
+
+use crate::message::CIDMap;
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+/// A simplified struct representing a `ChainHead` response that does not decode the `blocks` field.
+#[derive(Debug, Deserialize, Serialize)]
```
adlrocha, line 10: The plan is to move all types here, right? Would you mind opening an issue explaining what needs to be moved to the `sdk` in a follow-up PR. Thanks!

src/lotus/message/tests.rs
```diff
@@ -31,8 +31,8 @@ fn test_ipc_address_from_map() {
     {
         "From": {
             "SubnetId": {
-                "Parent": "/r123/f01",
-                "Actor": "f064"
+                "Root": "123",
```
adlrocha, line 34: Are you sure this is a string? I would have thought that this is a `u64` from the type of the field in `SubnetID`?

src/manager/evm/manager.rs
```diff
@@ -78,6 +80,36 @@ struct IPCContractInfo {
     provider: Provider<Http>,
 }
 
+#[async_trait]
+impl TopDownCheckpointQuery for EthSubnetManager {
+    async fn chain_head(&self) -> Result<ChainHeadResponse> {
+        unimplemented!()
```
adlrocha, line 86: Are you planning to implemented in the future, why is this unimplemented? (add a new issue if you are planning a follow-up and ideally point it from here). Thanks!

src/manager/fevm.rs
```diff
@@ -34,6 +56,168 @@ impl FevmSubnetManager {
...
+        self.top_down_msgs(subnet_id, start_epoch, end_epoch).await
+    }
+
+    async fn get_block_hash(&self, height: ChainEpoch) -> anyhow::Result<Vec<u8>> {
+        self.evm_subnet_manager.get_block_hash(height).await
+    }
+}
+
+#[async_trait]
+impl SubnetManager for FevmSubnetManager {
```
adlrocha, line 80: What has changed here? Why do we need to reimplement this one?

adlrocha (approved):
It was a bit hard to review this PR because it mixed code that was moved along with the implementation of new methods, so please bear with me in the comments. 

Also I think we should call this the `ipc-sdk` and we may need a huge refactor to improve it a bit. We can discuss it as a follow-up.

ipc-agent-sdk/Cargo.toml
cryptoAtwill: There is already a `ipc-sdk`, unless we merge them together.

src/manager/fevm.rs
cryptoAtwill: Previously it's `EvmSubnetManager`, this is `FevmSubnetManager`.

src/checkpoint/topdown.rs
```diff
@@ -108,43 +108,8 @@ impl<P: TopDownHandler, C: TopDownHandler> CheckpointManager for TopDownManager<
...
-
-        log::info!(
-            "checkpoint at epoch {:} for manager: {:} published with at epoch: {:?}, executed",
-            epoch,
-            self,
-            submitted_epoch,
-        );
-
-        Ok(())
+    async fn submit_checkpoint(&self, _epoch: ChainEpoch, _validator: &Address) -> Result<()> {
```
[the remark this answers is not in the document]
cryptoAtwill: yeah, it breaks the trait... I feel we might need to do some house cleaning with the new checkpoint mechanism.

src/manager/fvm/mod.rs
```diff
@@ -45,6 +47,26 @@ pub struct LotusSubnetManager<T: JsonRpcClient> {
     gateway_addr: Address,
 }
 
+#[async_trait]
+impl<T: JsonRpcClient + Send + Sync> TopDownCheckpointQuery for LotusSubnetManager<T> {
+    async fn chain_head(&self) -> Result<ChainHeadResponse> {
+        self.lotus_client.chain_head().await
+    }
+
+    async fn get_top_down_msgs(
```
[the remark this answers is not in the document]
cryptoAtwill: yeah, I think we should do it early.

src/manager/fevm.rs
adlrocha: Why do we need both of them? I think we can remove one of them because they're redundant, right?

adlrocha (commented):
Do you mind fixing conflicts and having a look at my last comment to get this one merged? Thanks!

src/server/handlers/manager/topdown_msgs.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+impl TopDownMsgsHandler {
+    pub(crate) fn new(pool: Arc<SubnetManagerPool>) -> Self {
+        Self { pool }
+    }
+}
+
+#[async_trait]
+impl JsonRPCRequestHandler for TopDownMsgsHandler {
```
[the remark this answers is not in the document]
cryptoAtwill: I think the current `sdk` is more or less doing this? One can just pass the url and token to it and it will point to the correct `ipc-agent` if that's what you mean.

ipc-agent-sdk/src/apis/topdown.rs
cryptoAtwill: Added

ipc-agent-sdk/src/message/chain.rs
cryptoAtwill: Added

src/manager/evm/manager.rs
cryptoAtwill: It's kind of difficult to do as Evm does not necessarily have all the types in `ChainHeadResponse`. Come to think of it, it does not make sense to have this trait implementation. I have removed this implementation.

cryptoAtwill: It's kind of difficult to do as Evm does not necessarily have all the types in `ChainHeadResponse`. Come to think of it, it does not make sense to have this trait implementation. I have updated the trait definition to return chain epoch as this is what top down checkpoint in fendermint needs.

src/manager/fevm.rs
cryptoAtwill: RIght, this is because `evm` and `fevm` are different with existing checkpoint mechanism. Because `fevm` checkpoint bottom up proof requires CID type. But once checkpoint in `fendermint` is merged, we should do a clean up on existing code base. Here is the issue I created to track this: https://github.com/consensus-shipyard/ipc-agent/issues/297

src/manager/fvm/mod.rs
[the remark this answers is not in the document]
cryptoAtwill: Created issue to track this: https://github.com/consensus-shipyard/ipc-agent/issues/297

src/lotus/message/tests.rs
cryptoAtwill: actually the deserialization can handle both cases, but I changed them to number.

PR #300, opened 2023-08-28
IPC-299: Groundwork for the refactor

adlrocha:
Feature branch for #299. All the PRs will use this branch as their base branch. In the next few PRs the following will be implemented: 
- [ ] Migration of `ipc-sdk`
- [ ] Implementation of the `ipc-provider` as a standalone library to interact with subnets leveraging the `sdk` and `manager` implementations.
- [ ] Implementation of the `ipc-cli` leveraging `ipc-provider` with the relevant commands to call from the cli leveraging the `ipc-provider` 
- [ ] Implementation of the `ipc-daemon` exposing the `agent` RPC server and becoming the main process responsible for spawning all the background processes of IPC. 

--------------
### Initial description before the this PR became the feature branch
This PR prepares the repo for some of the refactor by: 
- Moving all subcrates into an `ipc` directory that will hold all the relevant packages for ipc. 
- Rename `ipc-agent-sdk` to `ipc-provider`. It will become the provider or client of the IPC agent.
-  Move all code for the agent into `ipc/agent`. This package will be moved around into the different subcrates and will be eventually become the `daemon`.


cryptoAtwill:
I think this is looking good. Just have one question on where to put the cli. I feel it can be moved out of `agent`. In agent, are we just providing a `json rpc` api, or how do we draw the boundary?

adlrocha:
> I think this is looking good. Just have one question on where to put the cli. I feel it can be moved out of agent. In agent, are we just providing a json rpc api, or how do we draw the boundary?

It will be its own crate, I am already working on that. Actually, all of this will make more sense once I wrap up the next PR (I can directly push it here and update the description accordingly instead of doing a new PR if it is easier to review for you).

PR #305, opened 2023-09-06
IPC-299: IpcProvider implementation

adlrocha:
This PR is the beginning of a sequence of them involving the core of the agent's refactor:
* Implements the skeleton of a new `IpcProvider` library that can be integrated to interact with IPC subnets.


cryptoAtwill (approved):

ipc/provider/src/lib.rs
```diff
@@ -4,8 +4,407 @@
...
+    ) -> anyhow::Result<Address> {
+        let conn = match self.connection(parent) {
+            None => return Err(anyhow!("target parent subnet not found")),
+            Some(conn) => conn,
+        };
+
+        let subnet_config = conn.subnet();
+        self.check_subnet(subnet_config)?;
+
+        let constructor_params = ConstructParams {
```
cryptoAtwill, line 187: btw, should we update the params here, a lot of the fields are outdated.

adlrocha: Should we update once we integrate with the contracts? I am still testing against the current version of contracts currently included in `main` and in the agent.

adlrocha: I am also thinking about propagating the compilation of actors in compile time that Akosh is using in Fendermint here so we don't have to worry about version mismatches.

PR #306, opened 2023-09-06
IPC-299: Set default key in EVM keystore and improvements

adlrocha:
This PR includes changes to the EVM keystore to improve the usability and allow the use of a default key. This is a requirement for the proper integration of the `IpcProvider` (see #305).

However, the keystore and identity crate is still a mess (along with the dichotomy betweem f4 and 0x addresses). Improvements to the current dirty fixes from this PR will be implemented in:
* https://github.com/consensus-shipyard/ipc-agent/issues/282
* https://github.com/consensus-shipyard/ipc-agent/issues/308


ipc/identity/src/evm/mod.rs
```diff
@@ -53,6 +57,12 @@ impl Drop for KeyInfo {
     }
 }
 
+/// This trait is use to determine the key chosen for a specific
+/// key in a general way.
+pub trait Defaultable {
```
cryptoAtwill, line 62: My first impression was `Default` for general types, but it happens to be about default key. Also returning `Self` implies `Self` is a `key`?

ipc/identity/src/evm/mod.rs
```diff
@@ -69,3 +79,50 @@ pub fn random_key_info() -> KeyInfo {
...
+}
+
+#[cfg(feature = "with-ethers")]
+impl ToString for EthKeyAddress {
+    fn to_string(&self) -> String {
+        self.inner.to_string()
+    }
+}
+
+#[cfg(feature = "with-ethers")]
```
cryptoAtwill, line 119: maybe we can put all the `#[cfg(feature = "with-ethers")]` code into a separate module/file, then we just enable `#[cfg(feature = "with-ethers")]` in mod file.

ipc/identity/src/evm/persistent.rs
```diff
@@ -16,8 +16,10 @@ use std::io::{BufReader, BufWriter, ErrorKind};
 use std::path::PathBuf;
 use zeroize::Zeroize;
 
+use super::Defaultable;
+
 #[derive(Default)]
-pub struct PersistentKeyStore<T> {
+pub struct PersistentKeyStore<T: Defaultable + ToString> {
```
cryptoAtwill, line 22: maybe if we dont need `T: Defaultable + ToString` in the struct definition, so that some `Impl` do not need to include `impl<T: Clone + Eq + Hash + TryFrom<KeyInfo> + Defaultable + ToString>`

cryptoAtwill (approved):

ipc/identity/src/evm/persistent.rs
adlrocha: fixed. Ended up using `Default` instead

ipc/identity/src/evm/mod.rs
adlrocha: fixed.

ipc/identity/src/evm/mod.rs
adlrocha: I'll probably leave this one for the next refactor if you don't mind so I don't mess up this PR more.

PR #307, opened 2023-09-06
IPC-299: Implement wallet CLI commands and provider identity

adlrocha:
This PR implements the `wallet` cli commands and the `subnet create` to de-risk the ability to use a daemon-less provider to interact with subnets. 

The API of `ipc_identity` and `ipc_provider` are still a mess, so don´t pay much attention to them. My goal in the next few PRs is to organize all the code, implement all the functions and cli command needed (except the cross-net messages) in order to start showcasing them at Fil Dev Summit, and then do a proper design for the API of the provider and the identity module.


ipc/cli/src/commands/wallet/balances.rs
```diff
@@ -21,26 +22,66 @@ impl CommandLineHandler for WalletBalances {
...
 
-        let url = get_ipc_agent_url(&arguments.ipc_agent_url, global)?;
-        let json_rpc_client = JsonRpcClientImpl::new(url, None);
+        let provider = get_ipc_provider(global)?;
 
-        let params = WalletBalancesParams {
-            subnet: arguments.subnet.clone(),
-        };
+        let wallet_type = WalletType::from_str(&arguments.wallet_type)?;
+        let subnet = Arc::new(SubnetID::from_str(&arguments.subnet)?);
```
cryptoAtwill, line 28: We might be able to remove the `Arc` and use reference and we dont have to clone.

ipc/cli/src/commands/wallet/balances.rs
```diff
@@ -21,26 +22,66 @@ impl CommandLineHandler for WalletBalances {
...
+        match wallet_type {
+            WalletType::Evm => {
+                let wallet = provider.evm_wallet();
+                let addresses = wallet.read().unwrap().list()?;
+                let r = addresses
+                    .iter()
+                    .map(|addr| {
+                        let provider = provider.clone();
+                        let subnet = subnet.clone();
+                        async move {
```
cryptoAtwill, line 38: do we need async `move`? Seems like `provider` can be Send + Sync, then we dont have to clone.

ipc/provider/src/lib.rs
```diff
@@ -149,11 +168,53 @@ impl IpcProvider {
...
             config::subnet::SubnetConfig::Fevm(_) => {
-                // TODO: add more checks later
+                // TODO: More checks to come
             }
         }
         Ok(())
     }
+
+    fn check_sender(
+        &mut self,
```
cryptoAtwill, line 178: making this `mut` will lead to methods that require sending `txn` to use `mut self`, which might force owner of this struct to use `mut self`, i.e. `fendermint` top down provider might encounter this issue. If `mut` is not necessary, we can just stick with `self`, else callers might have to use `RWLock` for this.

cryptoAtwill (approved):

ipc/cli/src/commands/wallet/balances.rs
adlrocha: Unfortunately, the compiler insists that I use `move` because if not some values like `addr` and `provider` will be outlived due to the loop :(

ipc/cli/src/commands/wallet/balances.rs
adlrocha: fixed

ipc/provider/src/lib.rs
adlrocha: Tracked in this issue for a follow-up: https://github.com/consensus-shipyard/ipc-agent/issues/315

PR #309, opened 2023-09-08
IPC-299: Implementation of provider methods

adlrocha:
This PR includes the implementation of all the relevant methods currently required for the cli in the `IpcProvider`.


ipc/provider/src/lib.rs
```diff
@@ -263,63 +263,144 @@ impl IpcProvider {
...
 
-    /// Lists all the registered children in a gateway.
     pub async fn list_child_subnets(
         &self,
-        _gateway_addr: Address,
+        gateway_addr: Option<Address>,
+        subnet: &SubnetID,
     ) -> anyhow::Result<HashMap<SubnetID, SubnetInfo>> {
-        todo!()
+        let parent = subnet.parent().ok_or_else(|| anyhow!("no parent found"))?;
```
cryptoAtwill, line 334: probably we dont need this check as we are preventing `root` subnets to be listed

ipc/provider/src/lib.rs
```diff
@@ -263,63 +263,144 @@ impl IpcProvider {
...
+
+        let subnet_config = conn.subnet();
+        self.check_subnet(subnet_config)?;
+
+        let gateway_addr = match gateway_addr {
+            None => subnet_config.gateway_addr(),
+            Some(addr) => addr,
+        };
+
+        conn.manager().list_child_subnets(gateway_addr).await
```
cryptoAtwill, line 348: Gateway address should not be needed in `list_child_subnets` as `manager` should already know which subnet it is `managing` and hence should have already known the `gateway` address. Ideally, manager should have owned the `gateway` address?.

ipc/provider/src/lib.rs
```diff
@@ -263,63 +263,144 @@ impl IpcProvider {
...
-        _amount: TokenAmount,
+        &mut self,
+        subnet: SubnetID,
+        gateway_addr: Option<Address>,
+        from: Option<Address>,
+        to: Address,
+        amount: TokenAmount,
     ) -> anyhow::Result<ChainEpoch> {
-        todo!()
+        let parent = subnet.parent().ok_or_else(|| anyhow!("no parent found"))?;
```
cryptoAtwill, line 359: maybe we should remove this as well, so that we can fund address in `root` subnets

cryptoAtwill (commented):
I feel we might not need `gateway` address as parameters anymore since it should be part of the subnet config already and we should use the one passed in from config.

ipc/provider/src/lib.rs
adlrocha: Good catch. Typo due to copy-pasting. Fixed.

ipc/provider/src/lib.rs
adlrocha: I don't think this is the case, right? When funding you need to connect to the parent and send a `fund` to the gateway. So there shouldn't be a `fund` from the root as an argument.

ipc/provider/src/lib.rs
adlrocha: As this is a huge change, I will fix it in a follow up PR. Tracked in this issue: https://github.com/consensus-shipyard/ipc-agent/issues/316

PR #310, opened 2023-09-08
IPC-299: Implementation of subnet cli commands

adlrocha:
This PR switches the implementation of the `subnet` commands of the cli to use the `IpcProvider` instead of the daemon.


ipc/cli/src/commands/subnet/list_validators.rs
```diff
@@ -23,23 +20,20 @@ impl CommandLineHandler for ListValidators {
...
 
-        let url = get_ipc_agent_url(&arguments.ipc_agent_url, global)?;
-        let json_rpc_client = JsonRpcClientImpl::new(url, None);
+        let provider = get_ipc_provider(global)?;
+        let subnet = SubnetID::from_str(&arguments.subnet)?;
 
-        let params = QueryValidatorSetParams {
-            subnet: arguments.subnet.clone(),
-            epoch: None,
+        let gateway_addr = match &arguments.gateway_address {
```
cryptoAtwill, line 26: same as previous PR, I think we dont need gateway address anymore

cryptoAtwill (approved):

ipc/cli/src/commands/subnet/list_validators.rs
adlrocha: Tracked in #316

PR #311, opened 2023-09-08
IPC-299: Implement cli cross-msg methods

adlrocha:
This PR implements the `cross` cli methods leveraging the `IpcProvider`


cryptoAtwill (commented):
We can use `parser` to handle the command line args to directly parse into the corresponding types: https://github.com/consensus-shipyard/fendermint/blob/4b6eba890764f87ed9d2947201822037d00054f9/fendermint/app/src/options/genesis.rs#L69

PR #318, opened 2023-09-14
IPC-314: Remove unnecessary code for daemon, and fvm-related

### LINKED ISSUE #314: Remove `fvm` manager, `daemon and unnecessary type conversions
This is part of #299. The idea is to remove all the crates and modules that are no longer necessary, and focus on what we want to support now which are Fendermint-based (and maybe Lotus-based) IPC subnets using the EVM runtime.

adlrocha:
Closes #314

This PR removes: 
* The daemon crate
* Unnecessary dependencies
* The FVM manager
* All checkpoint-specific logic, as it is changing a lot and it will be included inside Fendermint

_Note: There may still be some unnecessary type conversions that I may have left behind and that may be removed in future iterations_.


cryptoAtwill:
@adlrocha we need also remove the unused `crates` in `Cargo.toml`, like `ipc-gateway` and `subnet-actor` crates in provider crates' Cargo.toml.

cryptoAtwill (approved):

PR #325, opened 2023-09-20
migrate abi

cryptoAtwill:
Migrate the abi using the new `ipc-actor-abis` crate. 

Need https://github.com/consensus-shipyard/ipc-solidity-actors/pull/205 to be merged first for the ci to pass.


ipc/provider/src/config/serialize.rs
```diff
@@ -98,16 +98,6 @@ mod tests {
     [server]
     json_rpc_address = "127.0.0.1:3030"
 
-    [[subnets]]
-    id = "/r123"
-    network_name = "test"
-
-    [subnets.config]
-    network_type = "fvm"
```
adlrocha, line 106: I was planning to leave the FVM config for now, but I guess you are right and is better just to remove it

adlrocha (approved):
Good one! Now we have consistent bindings in all the stack. Thanks

ipc/provider/src/config/serialize.rs
cryptoAtwill: yeah, it's breaking the tests and seems no point keeping.

PR #327, opened 2023-10-03
add query validators

cryptoAtwill:
Support querying validator change set for updated top down parent finality. Need also to remove some code as abi was updated drastically.


ipc/provider/src/manager/subnet.rs
```diff
@@ -141,4 +142,11 @@ pub trait TopDownCheckpointQuery: Send + Sync {
     ) -> Result<Vec<CrossMsg>>;
     /// Get the block hash
     async fn get_block_hash(&self, height: ChainEpoch) -> Result<Vec<u8>>;
+    /// Get the validator change set from start to end block.
```
adlrocha, line 145: Do you mind being more explicit? We expect the `IpcProvider` to be used by application. It may be useful to explain that this returns all the changes on the membership between two blocks. "validator change set" seems like an opaque concept.

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -485,16 +512,8 @@ impl SubnetManager for EthSubnetManager {
...
-        self.ensure_same_gateway(gateway_addr)?;
-
-        let gateway_contract = gateway_getter_facet::GatewayGetterFacet::new(
-            self.ipc_contract_info.gateway_addr,
-            Arc::new(self.ipc_contract_info.provider.clone()),
-        );
-        let epoch = gateway_contract.last_voting_executed_epoch().call().await?;
-
-        Ok(epoch as ChainEpoch)
+    async fn last_topdown_executed(&self, _gateway_addr: &Address) -> Result<ChainEpoch> {
```
adlrocha, line 515: Does this still exists with top-down proofs of finality? If this is the case we may want to rename the function.

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -102,6 +102,34 @@ impl TopDownCheckpointQuery for EthSubnetManager {
...
+    ) -> Result<Vec<StakingChangeRequest>> {
+        let address = contract_address_from_subnet(subnet_id)?;
+        log::info!("querying validator changes in evm subnet contract: {address:}");
+
+        let contract = subnet_actor_manager_facet::SubnetActorManagerFacet::new(
+            address,
+            Arc::new(self.ipc_contract_info.provider.clone()),
+        );
+
+        let ev = contract
```
adlrocha, line 120: So here you are getting all events that trigger __any__ change in membership, right?

ipc/sdk/src/lib.rs
```diff
@@ -88,3 +91,17 @@ impl ValidatorSet {
         self.configuration_number += 1;
     }
 }
+
+/// Converts an ethers::U256 TokenAmount into a FIL amount.
```
adlrocha, line 95: Out of a curiosity, why did you move these methods?

adlrocha (approved):

PR #331, opened 2023-10-05
Support top down query

cryptoAtwill:
Add more top down finality util queries in provider.

### DIFF
--- a/ipc/provider/src/lib.rs
+++ b/ipc/provider/src/lib.rs
@@ -535,6 +535,7 @@ impl IpcProvider {
         conn.manager().chain_head_height().await
     }
 
+    /// Obtain the genesis epoch of the input subnet.
     pub async fn genesis_epoch(&self, subnet: &SubnetID) -> anyhow::Result<ChainEpoch> {
         let parent = subnet.parent().ok_or_else(|| anyhow!("no parent found"))?;
         let conn = match self.connection(&parent) {
@@ -544,6 +545,7 @@ impl IpcProvider {
         conn.manager().genesis_epoch(subnet).await
     }
 
+    /// Get the changes in subnet validators. This is fetched from parent.
     pub async fn get_validator_changeset(
         &self,
         subnet: &SubnetID,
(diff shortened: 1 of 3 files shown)


ipc/provider/src/lib.rs
```diff
@@ -572,6 +598,25 @@ impl IpcProvider {
...
+    pub async fn get_chain_head_height(&self, subnet: &SubnetID) -> anyhow::Result<ChainEpoch> {
+        let conn = match self.connection(subnet) {
+            None => return Err(anyhow!("target subnet not found")),
+            Some(conn) => conn,
+        };
+
+        conn.manager().chain_head_height().await
+    }
+
+    pub async fn get_genesis_epoch(&self, subnet: &SubnetID) -> anyhow::Result<ChainEpoch> {
```
adlrocha, line 611: Is this a duplicate of the `genesis_epoch` above?

ipc/provider/src/lib.rs
```diff
@@ -572,6 +598,25 @@ impl IpcProvider {
...
+        let parent = subnet.parent().ok_or_else(|| anyhow!("no parent found"))?;
+        let conn = match self.connection(&parent) {
+            None => return Err(anyhow!("parent subnet not found")),
+            Some(conn) => conn,
+        };
+
+        conn.manager().chain_head_height().await
```
adlrocha, line 618: Also, this doesn't seem what you want to do, right? :)

ipc/provider/src/lib.rs
```diff
@@ -534,6 +535,31 @@ impl IpcProvider {
...
+        start: ChainEpoch,
+        end: ChainEpoch,
+    ) -> anyhow::Result<Vec<StakingChangeRequest>> {
+        let conn = match self.connection(subnet) {
+            None => return Err(anyhow!("target subnet not found")),
+            Some(conn) => conn,
+        };
+
+        conn.manager()
+            .get_validator_changeset(subnet, start, end)
```
adlrocha, line 559: Where is this one implemented?

ipc/provider/src/manager/subnet.rs
```diff
@@ -128,6 +128,8 @@ pub trait SubnetManager: Send + Sync + TopDownCheckpointQuery {
 /// Trait to interact with a subnet to query the necessary information for top down checkpoint.
 #[async_trait]
 pub trait TopDownCheckpointQuery: Send + Sync {
+    /// Returns the chain head height
```
adlrocha, line 131: This comment is wrong, right?

ipc/provider/src/lib.rs
```diff
@@ -572,6 +598,25 @@ impl IpcProvider {
 
         conn.manager().get_chain_id().await
     }
+
+    pub async fn get_chain_head_height(&self, subnet: &SubnetID) -> anyhow::Result<ChainEpoch> {
+        let conn = match self.connection(subnet) {
```
adlrocha, line 603: Isn't the change set picked up from the parent?

adlrocha (changes requested):
There seems to be a lot of copy-pasta and things missing (or maybe I am looking at the wrong commit?)

ipc/provider/src/lib.rs
cryptoAtwill: this is get chain head height? Anyways, for change set, it's from the subnet actor in parent, so the address should be derived from the subnet directly.

ipc/provider/src/lib.rs
cryptoAtwill: right, this method should be removed.

adlrocha (approved):

PR #335, opened 2023-10-13
Detect chain reorg

cryptoAtwill:
Changes for detecting chain reorgs.

### DIFF
--- a/ipc/provider/src/manager/evm/manager.rs
+++ b/ipc/provider/src/manager/evm/manager.rs
@@ -24,7 +24,7 @@ use ethers::prelude::k256::ecdsa::SigningKey;
 use ethers::prelude::{Signer, SignerMiddleware};
 use ethers::providers::{Authorization, Http, Middleware, Provider};
 use ethers::signers::{LocalWallet, Wallet};
-use ethers::types::{Eip1559TransactionRequest, I256, U256};
+use ethers::types::{BlockId, Eip1559TransactionRequest, I256, U256};
 use futures_util::StreamExt;
 use fvm_shared::clock::ChainEpoch;
 use fvm_shared::{address::Address, econ::TokenAmount};
@@ -113,15 +113,11 @@ impl TopDownCheckpointQuery for EthSubnetManager {
             self.ipc_contract_info.gateway_addr,
             Arc::new(self.ipc_contract_info.provider.clone()),
         );
-        let (raw_msgs, hash) = gateway_contract
-            .method::<_, (Vec<gateway_getter_facet::CrossMsg>, [u8; 32])>(
-                "getTopDownMsgs",
-                gateway_getter_facet::GetTopDownMsgsCall {
-                    subnet_id,
-                    block_number: U256::from(epoch),
-                },
-            )
-            .map_err(|e| anyhow!("cannot create the top down msg call: {e:}"))?
+
+        let call = gateway_contract
+            .get_top_down_msgs(subnet_id, U256::from(epoch))
+            .block(BlockId::from(epoch as u64));
+        let (raw_msgs, hash) = call
             .call()
             .await
             .map_err(|e| anyhow!("cannot get evm top down messages: {e:}"))?;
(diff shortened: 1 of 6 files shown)


ipc/provider/src/manager/evm/manager.rs
```diff
@@ -100,32 +100,65 @@ impl TopDownCheckpointQuery for EthSubnetManager {
...
+            Arc::new(self.ipc_contract_info.provider.clone()),
+        );
+        let (raw_msgs, hash) = gateway_contract
+            .method::<_, (Vec<gateway_getter_facet::CrossMsg>, [u8; 32])>(
+                "getTopDownMsgs",
+                gateway_getter_facet::GetTopDownMsgsCall {
+                    subnet_id,
+                    block_number: U256::from(epoch),
+                },
+            )
```
aakoshh, line 123: Why don't you use the statically typed ABI method?

cryptoAtwill: More of a legacy issue. The idea is we should query up till the target block, i.e. say if we want to get the messages up till block 100, but current head is 200. We still want to use the state at block 100 so that everyone has the same view. Back then this way was tested and works. It seems using the statically typed ABI could also support using block number state, but we have not tested it yet.

aakoshh: Like [here](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/eth/api/examples/ethers.rs#L541-L548). I'm curious because if you did, you wouldn't have to declare the return type. The compiler would tell you if there is any mismatch too.

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -136,16 +169,27 @@ impl TopDownCheckpointQuery for EthSubnetManager {
...
+            } else {
+                hash = Some(meta.block_hash);
+            }
             changes.push(StakingChangeRequest::try_from(event)?);
         }
 
-        Ok(changes)
+        Ok(TopDownQueryPayload {
+            value: changes,
+            block_hash: hash.unwrap().0.to_vec(),
```
aakoshh, line 191: What if there were no events at that block height? It looks to me like `hash` would stay empty and this `.unwrap()` would panic.

cryptoAtwill: yep, you are right. In this case, I should have changed the return type to: `Option<TopDownQueryPayload<Vec<StakingChangeRequest>>>`

adlrocha (approved):
Feel free to merge when you address @aakoshh comments :pray:

ipc/provider/src/manager/evm/manager.rs
cryptoAtwill: Yeah, we can give it a try. Should work.

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -100,32 +100,65 @@ impl TopDownCheckpointQuery for EthSubnetManager {
...
+        let (raw_msgs, hash) = gateway_contract
+            .method::<_, (Vec<gateway_getter_facet::CrossMsg>, [u8; 32])>(
+                "getTopDownMsgs",
+                gateway_getter_facet::GetTopDownMsgsCall {
+                    subnet_id,
+                    block_number: U256::from(epoch),
+                },
+            )
+            .map_err(|e| anyhow!("cannot create the top down msg call: {e:}"))?
+            .call()
```
aakoshh, line 125: For reference why I don't think this won't work: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/240/files#r1358015064

That said, there is an option to set the `BlockId` of a `ContractCall` before it's sent, but I have no idea what would happen in the EVM:
* currently we send a call to `getTopDownMsgs(height)` at the latest block height, so if `height < latest.height - 256` this will not work
* we could _also_ set the `height` on the `ContractCall` itself, so that _relatively_ it wouldn't be trailing the queried block by 256, it would match, but the EVM might not care

aakoshh: The EVM actor compares always to the [curr_epoch](https://github.com/filecoin-project/builtin-actors/blob/master/actors/evm/src/interpreter/instructions/context.rs#L25) of the runtime. That might be initialized to where the query was run, at least that's what I do in Fendermint, not sure about Lotus.

aakoshh (approved):
Cheers :)

PR #337, opened 2023-10-16
FM-306: add method in provider to get genesis info

adlrocha:
Related to https://github.com/consensus-shipyard/fendermint/issues/306
Depends on https://github.com/consensus-shipyard/ipc-agent/pull/336

This PR introduces a new method in the provider to retrieve all the genesis information in the parent required to deterministically generate the genesis for a Fendermint child subnet.

It also introduces `min_cross_msg_fee` as an additional configuration parameter that users can set when they create their subnets and many other changes.

### DIFF
--- a/ipc/provider/src/manager/evm/manager.rs
+++ b/ipc/provider/src/manager/evm/manager.rs
@@ -11,6 +11,7 @@ use ipc_actors_abis::{
     subnet_actor_getter_facet, subnet_actor_manager_facet, subnet_registry,
 };
 use ipc_sdk::evm::{fil_to_eth_amount, payload_to_evm_address, subnet_id_to_evm_addresses};
+use ipc_sdk::validator::from_contract_validators;
 use ipc_sdk::{eth_to_fil_amount, ethers_address_to_fil_address};
 
 use crate::config::subnet::SubnetConfig;
@@ -216,7 +217,7 @@ impl SubnetManager for EthSubnetManager {
             .min_cross_msg_fee
             .atto()
             .to_u128()
-            .ok_or_else(|| anyhow!("invalid min validator stake"))?;
+            .ok_or_else(|| anyhow!("invalid min message fee"))?;
 
         log::debug!("calling create subnet for EVM manager");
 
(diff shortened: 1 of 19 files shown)


ipc/provider/src/manager/subnet.rs
```diff
@@ -111,6 +112,20 @@ pub trait SubnetManager: Send + Sync + TopDownCheckpointQuery {
...
+
+#[derive(Debug)]
+pub struct SubnetGenesisInfo {
+    pub bottom_up_checkpoint_period: u64,
+    pub msg_fee: TokenAmount,
+    pub majority_percentage: u8,
+    pub active_validators_limit: u16,
+    pub min_collateral: TokenAmount,
+    pub genesis_epoch: ChainEpoch,
+    pub validators: Vec<subnet_actor_getter_facet::Validator>,
```
cryptoAtwill, line 128: we should use our own definition of `Validator`. Using `subnet_actor_getter_facet::Validator` will be carried to other functions or contract calls or in the future other network types. It might haunt us in the futures.

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -596,6 +605,29 @@ impl SubnetManager for EthSubnetManager {
...
+
+    async fn get_genesis_info(&self, subnet: &SubnetID) -> Result<SubnetGenesisInfo> {
+        let address = contract_address_from_subnet(subnet)?;
+        let contract = subnet_actor_getter_facet::SubnetActorGetterFacet::new(
+            address,
+            Arc::new(self.ipc_contract_info.provider.clone()),
+        );
+        Ok(SubnetGenesisInfo {
+            // Active validators limit set for the child subnet.
+            active_validators_limit: contract.active_validators_limit().call().await?,
```
cryptoAtwill, line 617: I think it needs some type conversion first?

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -207,6 +209,12 @@ impl SubnetManager for EthSubnetManager {
             .to_u128()
             .ok_or_else(|| anyhow!("invalid min validator stake"))?;
 
+        let min_cross_msg_fee = params
```
cryptoAtwill, line 212: This should be added in dev.

cryptoAtwill (changes requested):

ipc/provider/src/manager/evm/manager.rs
adlrocha: I don't think so, it returns the right type right away. I tested it and it works.

ipc/provider/src/manager/evm/manager.rs
adlrocha: Yes, I will solve the merge conflicts in a minute.

cryptoAtwill (approved):

PR #338, opened 2023-10-17
Bottom up checkpoint daemon

cryptoAtwill:
Adding bottom up checkpoint daemon as a struct. Entrypoint should be `BottomUpCheckpointManager`.


ipc/provider/src/checkpoint.rs
```diff
@@ -0,0 +1,143 @@
...
+use crate::manager::{BottomUpCheckpointRelayer, EthSubnetManager};
+use anyhow::{anyhow, Result};
+use fvm_shared::address::Address;
+use fvm_shared::clock::ChainEpoch;
+use ipc_identity::{EthKeyAddress, PersistentKeyStore};
+use std::fmt::{Display, Formatter};
+use std::sync::{Arc, RwLock};
+use std::time::Duration;
+
+/// Tracks the metadata required for both top down and bottom up checkpoint submissions, such as
```
adlrocha, line 15: Is this a typo? 
```suggestion
/// Tracks the metadata required for bottom up checkpoint submissions
```

ipc/provider/src/checkpoint.rs
```diff
@@ -0,0 +1,143 @@
...
+/// parent/child subnet and checkpoint period.
+pub struct CheckpointMetadata {
```
adlrocha, line 17: It is really checkpoint metadata or the config for the bottom-up checkpoint?
```suggestion
pub struct CheckpointConfig {
```

ipc/provider/src/checkpoint.rs
```diff
@@ -0,0 +1,143 @@
...
+    parent: Subnet,
+    child: Subnet,
+    period: ChainEpoch,
+}
+
+pub struct BottomUpCheckpointManager<P, C> {
```
adlrocha, line 23: I don't understand why you need two generics here `P` and `C` of type `CheckpointRelayer`. Wouldn't it be cleaner to have a single type `R` that is of type `CheckpointRelayer`. A relayer is always from a child to a parent. I don't think this abstraction is correct.

ipc/provider/src/checkpoint.rs
```diff
@@ -0,0 +1,143 @@
...
+    metadata: CheckpointMetadata,
+    parent_handler: P,
+    child_handler: C,
+}
+
+impl<P: BottomUpCheckpointRelayer, C: BottomUpCheckpointRelayer> BottomUpCheckpointManager<P, C> {
+    pub async fn new(
+        parent: Subnet,
+        child: Subnet,
+        parent_handler: P,
```
adlrocha, line 33: What you could maybe do here is to initialize an `IpcProvider` with access to the parent and the child and that implements the operations that you need for the relayer

ipc/provider/src/checkpoint.rs
```diff
@@ -0,0 +1,143 @@
...
+        write!(
+            f,
+            "bottom-up, parent: {:}, child: {:}",
+            self.metadata.parent.id, self.metadata.child.id
+        )
+    }
+}
+
+impl<
+        P: BottomUpCheckpointRelayer + Send + Sync + 'static,
```
adlrocha, line 78: As mentioned above, I would try to have a single generic abstraction for the parent-child relay

ipc/provider/src/checkpoint.rs
```diff
@@ -0,0 +1,143 @@
...
+        self.parent_handler
+            .submit_checkpoint(validator, bundle)
+            .await
+            .map_err(|e| anyhow!("cannot submit bottom up checkpoint due to: {e:}"))?;
+
+        Ok(())
+    }
+
+    async fn next_submission_height(&self) -> Result<ChainEpoch> {
+        let last_checkpoint_epoch = self
```
adlrocha, line 134: One caveat here, relayers are allowed to submit a checkpoint for the `lastBottomUpCheckpointHeight` (and rewarded if they do so) as long as they haven't submitted it yet, and the next checkpoint hasn't been submitted yet. Before the epoch of the next checkpoint, we should also check if we can submit the one for the current one  (even if someone has sent it already): https://github.com/consensus-shipyard/ipc-solidity-actors/blob/ad80d9b29c98473042c42b8f9cae5c7453620c25/src/subnet/SubnetActorManagerFacet.sol#L75

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -778,9 +783,123 @@ impl EthSubnetManager {
     }
 }
 
+#[async_trait]
+impl BottomUpCheckpointRelayer for EthSubnetManager {
+    async fn submit_checkpoint(
```
adlrocha, line 788: Should we make these checkpoint-related methods available through the `IpcProvider`?

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -778,9 +783,123 @@ impl EthSubnetManager {
...
+        );
+
+        let (checkpoint, _, signatories, signatures) =
+            contract.get_signature_bundle(height as u64).call().await?;
+        let cross_msgs = contract.bottom_up_messages(height as u64).call().await?;
+
+        let checkpoint = BottomUpCheckpoint::try_from(checkpoint)?;
+        let signatories = signatories
+            .into_iter()
+            .map(|s| ethers_address_to_fil_address(&s))
```
adlrocha, line 870: Why do we need to perform this translation into fil address?

ipc/provider/src/checkpoint.rs
cryptoAtwill: Because the parent and child will be different network type?

ipc/provider/src/manager/evm/manager.rs
cryptoAtwill: Because in `ipc-agent` the convention is to use `fev_shared::Address`? We are removing this?

ipc/provider/src/manager/evm/manager.rs
cryptoAtwill: I donno if that's required. But we can easily add them when we have to. I would say we keep it as it is, then expose them when needed to.

ipc/provider/src/checkpoint.rs
cryptoAtwill: As discussed over slack, there wont be a case for supporting different network type. So a single generic should do it.

ipc/provider/src/checkpoint.rs
cryptoAtwill: I think that works too, just I feel that way we are making both `IpcProvider` and `BottomUpCheckpoint` quite messy. `IpcProvider` is like a wrapper that coordinates and exposes `SubnetManager`, it just handles the connection and calls into `SubnetManager`. If we shift Checkpoint implementation to it, it will be more complex than other methods. We can ask `IpcProvider` to call into `BottomUpCheckpointManager` if needed. That way I think it's neater.

adlrocha: As discussed sync, let's keep the `IpcProvider` as a simple wrapper for external functions for now, and let's evolve it from user and DX feedback.

ipc/provider/src/manager/evm/manager.rs
adlrocha: Resolved in comment above.

adlrocha (approved):

PR #339, opened 2023-10-17
Relayer cli

cryptoAtwill:
Adding relayer cli plus some small fixes.

### DIFF
--- a/ipc/cli/src/commands/checkpoint/relayer.rs
+++ b/ipc/cli/src/commands/checkpoint/relayer.rs
@@ -37,7 +37,7 @@ impl CommandLineHandler for BottomUpRelayer {
                 Address::try_from(addr)?
             }
             _ => {
-                return Err(anyhow!("no validator address provided"));
+                return Err(anyhow!("no submitter address provided"));
             }
         };
 
--- a/ipc/cli/src/commands/checkpoint/relayer.rs
+++ b/ipc/cli/src/commands/checkpoint/relayer.rs
@@ -15,7 +15,7 @@ use std::str::FromStr;
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
 
-const DEFAULT_CHECKPOINT_INTERVAL: u64 = 15;
+const DEFAULT_POLLING_INTERVAL: u64 = 15;
 
 /// The command to run the bottom up relayer in the background.
 pub(crate) struct BottomUpRelayer;
@@ -30,8 +30,8 @@ impl CommandLineHandler for BottomUpRelayer {
         let config_path = global.config_path();
 
         let mut keystore = new_evm_keystore_from_path(&config_path)?;
-        let validator = match (arguments.validator.as_ref(), keystore.get_default()?) {
-            (Some(validator), _) => Address::from_str(validator)?,
+        let submitter = match (arguments.submitter.as_ref(), keystore.get_default()?) {
+            (Some(submitter), _) => Address::from_str(submitter)?,
             (None, Some(addr)) => {
                 log::info!("using default address: {addr:?}");
                 Address::try_from(addr)?
@@ -59,9 +59,9 @@ impl CommandLineHandler for BottomUpRelayer {
         let interval = Duration::from_secs(
             arguments
                 .checkpoint_interval_sec
-                .unwrap_or(DEFAULT_CHECKPOINT_INTERVAL),
+                .unwrap_or(DEFAULT_POLLING_INTERVAL),
         );
-        manager.run(validator, interval).await;
+        manager.run(submitter, interval).await;
 
         Ok(())
     }
@@ -74,6 +74,6 @@ pub(crate) struct BottomUpRelayerArgs {
     pub subnet: String,
     #[arg(long, short, help = "The number of seconds to submit checkpoint")]
     pub checkpoint_interval_sec: Option<u64>,
-    #[arg(long, short, help = "The hex encoded address of the validator")]
-    pub validator: Option<String>,
+    #[arg(long, short, help = "The hex encoded address of the submitter")]
+    pub submitter: Option<String>,
 }
--- a/ipc/cli/src/commands/checkpoint/relayer.rs
+++ b/ipc/cli/src/commands/checkpoint/relayer.rs
@@ -1,0 +1,79 @@
+// Copyright 2022-2023 Protocol Labs
+// SPDX-License-Identifier: MIT
+
+use crate::commands::get_subnet_config;
+use crate::{CommandLineHandler, GlobalArguments};
+use anyhow::anyhow;
+use async_trait::async_trait;
+use clap::Args;
+use fvm_shared::address::Address;
+use ipc_identity::EvmKeyStore;
+use ipc_provider::checkpoint::BottomUpCheckpointManager;
+use ipc_provider::new_evm_keystore_from_path;
+use ipc_sdk::subnet_id::SubnetID;
+use std::str::FromStr;
+use std::sync::{Arc, RwLock};
+use std::time::Duration;
+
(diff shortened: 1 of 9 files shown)


ipc/cli/src/commands/checkpoint/relayer.rs
```diff
@@ -0,0 +1,79 @@
...
+                .unwrap_or(DEFAULT_CHECKPOINT_INTERVAL),
+        );
+        manager.run(validator, interval).await;
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Args)]
+#[command(about = "Start the bottom up relayer")]
```
adlrocha, line 71:
```suggestion
#[command(about = "Start the bottom up relayer daemon")]
```

ipc/cli/src/commands/checkpoint/relayer.rs
```diff
@@ -0,0 +1,79 @@
...
+pub(crate) struct BottomUpRelayerArgs {
+    #[arg(long, short, help = "The subnet id of the checkpointing subnet")]
+    pub subnet: String,
+    #[arg(long, short, help = "The number of seconds to submit checkpoint")]
+    pub checkpoint_interval_sec: Option<u64>,
+    #[arg(long, short, help = "The hex encoded address of the validator")]
+    pub validator: Option<String>,
```
adlrocha, line 78: Wait, relayers don't need to be validators in the subnet, let's call this `address` as we do for the rest of the subnet.

I also don't understand why we need the `PubKey` here, I think we should pass the address and leverage the wallet to sign or any interactions required with the parent network.
```suggestion
    pub address: Option<String>,
```

ipc/cli/src/commands/checkpoint/relayer.rs
```diff
@@ -0,0 +1,79 @@
...
+    type Arguments = BottomUpRelayerArgs;
+
+    async fn handle(global: &GlobalArguments, arguments: &Self::Arguments) -> anyhow::Result<()> {
+        log::debug!("start bottom up relayer with args: {:?}", arguments);
+
+        let config_path = global.config_path();
+
+        let mut keystore = new_evm_keystore_from_path(&config_path)?;
+        let validator = match (arguments.validator.as_ref(), keystore.get_default()?) {
+            (Some(validator), _) => Address::from_str(validator)?,
```
adlrocha, line 34: As mentioned below, let's call this the `address`.

ipc/cli/src/commands/checkpoint/relayer.rs
```diff
@@ -0,0 +1,79 @@
...
+use fvm_shared::address::Address;
+use ipc_identity::EvmKeyStore;
+use ipc_provider::checkpoint::BottomUpCheckpointManager;
+use ipc_provider::new_evm_keystore_from_path;
+use ipc_sdk::subnet_id::SubnetID;
+use std::str::FromStr;
+use std::sync::{Arc, RwLock};
+use std::time::Duration;
+
+const DEFAULT_CHECKPOINT_INTERVAL: u64 = 15;
```
adlrocha, line 18: This is not the checkpoint interval but the polling time  in the process loop, right? I think we should rename it and clarify in a comment.
```suggestion
const DEFAULT_POLLING_INTERVAL: u64 = 15;
```

ipc/cli/src/commands/checkpoint/relayer.rs
adlrocha: I see, I think the public key side of things is actually a typo due to the copy-pasta, right?

ipc/cli/src/commands/checkpoint/relayer.rs
```diff
@@ -0,0 +1,79 @@
...
+            Arc::new(RwLock::new(keystore)),
+        )
+        .await?;
+
+        let interval = Duration::from_secs(
+            arguments
+                .checkpoint_interval_sec
+                .unwrap_or(DEFAULT_CHECKPOINT_INTERVAL),
+        );
+        manager.run(validator, interval).await;
```
adlrocha, line 64: Out of a curiosity, how are we handling killing the process? We should ensure that `ctrl-c` here gracefully kills the process.

ipc/provider/src/checkpoint.rs
```diff
@@ -89,20 +89,59 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
+                log::error!("cannot submit checkpoint for validator: {validator} due to {e}");
             }
-        });
+
+            tokio::time::sleep(submission_interval).await;
+        }
     }
 
     /// Submit the checkpoint from the target validator address
     pub async fn submit_checkpoint(&self, validator: &Address) -> Result<()> {
```
adlrocha, line 103: We should remove `validator` from all the functions as the argument (I think I missed it in my previous review). We don't need to be a `validator` to act as a relayer in the system.
```suggestion
    pub async fn submit_checkpoint(&self, relayer: &Address) -> Result<()> {
```

adlrocha (commented):
I left a few comments and suggestion. Something worth fixing in this PR and the previous is the use of `validator` for the `relayer`role. As mentioned we don't need to be validators to relay a checkpoint, and this may lead to confusion.

ipc/provider/src/checkpoint.rs
```diff
@@ -89,20 +89,59 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
+            .parent_handler
+            .last_bottom_up_checkpoint_height(&self.metadata.child.id)
+            .await
+            .map_err(|e| {
+                anyhow!("cannot obtain the last bottom up checkpoint height due to: {e:}")
+            })?;
+        Ok(last_checkpoint_epoch + self.checkpoint_period())
+    }
+
+    async fn submit_last_epoch(&self, validator: &Address) -> Result<()> {
```
adlrocha, line 119:
```suggestion
    /// Checks if the relayer has already submitted the last checkpoint, if not it submits it.
    async fn submit_last_epoch(&self, validator: &Address) -> Result<()> {
```

ipc/cli/src/commands/checkpoint/relayer.rs
cryptoAtwill: Yeah, ctrl+c should be able to kill the process.

ipc/cli/src/commands/checkpoint/relayer.rs
cryptoAtwill: I still have the impresseion that only validators can submit, I will change it to submitter

ipc/provider/src/checkpoint.rs
```diff
@@ -89,20 +92,62 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
     }
 
     /// Run the bottom up checkpoint submission daemon in the background
```
adlrocha, line 94: We are actually running the process in the foreground.
```suggestion
    /// Run the bottom up checkpoint submission daemon in the foreground
```

ipc/cli/src/commands/checkpoint/relayer.rs
```diff
@@ -0,0 +1,79 @@
...
+
+        let mut keystore = new_evm_keystore_from_path(&config_path)?;
+        let submitter = match (arguments.submitter.as_ref(), keystore.get_default()?) {
+            (Some(submitter), _) => Address::from_str(submitter)?,
+            (None, Some(addr)) => {
+                log::info!("using default address: {addr:?}");
+                Address::try_from(addr)?
+            }
+            _ => {
+                return Err(anyhow!("no validator address provided"));
```
adlrocha, line 40:
```suggestion
                return Err(anyhow!("no submitter address provided"));
```

adlrocha (approved):

PR #350, opened 2023-10-27
Query event

cryptoAtwill:
Query the quorum reached event to trigger processing of checkpoint submission.

### DIFF
--- a/ipc/cli/src/commands/checkpoint/bottomup_bundles.rs
+++ b/ipc/cli/src/commands/checkpoint/bottomup_bundles.rs
@@ -36,7 +36,7 @@ impl CommandLineHandler for GetBottomUpBundles {
 }
 
 #[derive(Debug, Args)]
-#[command(about = "List bottom up checkpoint bundles for a child subnet")]
+#[command(about = "List bottom up checkpoint signature bundle for a child subnet")]
 pub(crate) struct GetBottomUpBundlesArgs {
     #[arg(long, short, help = "The target subnet to perform query")]
     pub subnet: String,
--- a/ipc/cli/src/commands/checkpoint/bottomup_bundles.rs
+++ b/ipc/cli/src/commands/checkpoint/bottomup_bundles.rs
@@ -1,0 +1,47 @@
+// Copyright 2022-2023 Protocol Labs
+// SPDX-License-Identifier: MIT
+//! List bottom up bundles
+
+use std::fmt::Debug;
+use std::str::FromStr;
+
+use async_trait::async_trait;
+use clap::Args;
+use fvm_shared::clock::ChainEpoch;
+use ipc_sdk::subnet_id::SubnetID;
+
+use crate::commands::get_ipc_provider;
+use crate::{CommandLineHandler, GlobalArguments};
+
+/// The command to get bottom up bundles at height.
+pub(crate) struct GetBottomUpBundles;
+
+#[async_trait]
+impl CommandLineHandler for GetBottomUpBundles {
+    type Arguments = GetBottomUpBundlesArgs;
+
+    async fn handle(global: &GlobalArguments, arguments: &Self::Arguments) -> anyhow::Result<()> {
+        log::debug!("get bottom up bundles with args: {:?}", arguments);
+
+        let provider = get_ipc_provider(global)?;
+        let subnet = SubnetID::from_str(&arguments.subnet)?;
+
+        for h in arguments.from_epoch..=arguments.to_epoch {
+            let bundle = provider.get_bottom_up_bundle(&subnet, h).await?;
+            println!("{bundle:?}");
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Args)]
+#[command(about = "List bottom up checkpoint bundles for a child subnet")]
+pub(crate) struct GetBottomUpBundlesArgs {
+    #[arg(long, short, help = "The target subnet to perform query")]
+    pub subnet: String,
+    #[arg(long, short, help = "Include checkpoints from this epoch")]
+    pub from_epoch: ChainEpoch,
+    #[arg(long, short, help = "Include checkpoints up to this epoch")]
+    pub to_epoch: ChainEpoch,
+}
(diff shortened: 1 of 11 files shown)


ipc/cli/src/commands/checkpoint/bottomup_bundles.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    async fn handle(global: &GlobalArguments, arguments: &Self::Arguments) -> anyhow::Result<()> {
+        log::debug!("get bottom up bundles with args: {:?}", arguments);
+
+        let provider = get_ipc_provider(global)?;
+        let subnet = SubnetID::from_str(&arguments.subnet)?;
+
+        for h in arguments.from_epoch..=arguments.to_epoch {
+            let bundle = provider.get_bottom_up_bundle(&subnet, h).await?;
+            println!("{bundle:?}");
```
adlrocha, line 31: Would you mind making the print a bit prettier like I did for top-down cross-messages? The debug output is quite hard to read: https://github.com/consensus-shipyard/ipc/blob/65e92a324e4c3818ceeee90f1416d10d8ad9a107/ipc/cli/src/commands/crossmsg/topdown_cross.rs#L44

ipc/cli/src/commands/checkpoint/bottomup_height.rs
```diff
@@ -0,0 +1,42 @@
...
+
+        let height = provider.last_bottom_up_checkpoint_height(&subnet).await?;
+        println!("height: {height}");
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Args)]
+#[command(about = "Last bottom up checkpoint height for a child subnet")]
```
adlrocha, line 38: This is the last checkpoint committed in the parent, or the last height signed in the child? Is not clear to me. It would be good to clarify it.
```suggestion
#[command(about = "Last bottom up checkpoint height committed in a child subnet")]
```

ipc/cli/src/commands/checkpoint/bottomup_bundles.rs
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Args)]
+#[command(about = "List bottom up checkpoint bundles for a child subnet")]
```
adlrocha, line 39:
```suggestion
#[command(about = "List bottom up checkpoint signature bundle for a child subnet")]
```

ipc/cli/src/commands/checkpoint/bottomup_submitted.rs
```diff
@@ -0,0 +1,50 @@
...
+            .await?;
+        println!("has submitted: {submitted}");
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Args)]
+#[command(
+    about = "Check if the address has submitted in the last bottom up checkpoint of a child subnet"
```
adlrocha, line 43: This is a signature submission, right? Not the commitment of the bottom-up checkpoint in the parent.
```suggestion
    about = "Check if the address has submitted a signature in the last bottom up checkpoint of a child subnet"
```

ipc/cli/src/commands/checkpoint/relayer.rs
```diff
@@ -74,6 +80,12 @@ pub(crate) struct BottomUpRelayerArgs {
     pub subnet: String,
     #[arg(long, short, help = "The number of seconds to submit checkpoint")]
     pub checkpoint_interval_sec: Option<u64>,
+    #[arg(
```
adlrocha, line 83: This can generally be `1` or `2` as in Tendermint we don´t  expect re-orgs. That being said, this may come handy once we support other types of consensus algorithms in child subnets.

ipc/provider/src/manager/subnet.rs
```diff
@@ -205,6 +206,8 @@ pub trait BottomUpCheckpointRelayer: Send + Sync {
     async fn checkpoint_period(&self, subnet_id: &SubnetID) -> Result<ChainEpoch>;
     /// Get the checkpoint bundle at a specific height. If it does not exist, it will through error.
     async fn checkpoint_bundle_at(&self, height: ChainEpoch) -> Result<BottomUpCheckpointBundle>;
+    /// Queries the quorum reached events at target height.
```
adlrocha, line 209:
```suggestion
    /// Queries the signature quorum reached events at target height.
```

adlrocha (approved):
it would be good to clarify a bit some of the commands. But overall LGTM.

ipc/provider/src/checkpoint.rs
```diff
@@ -27,6 +28,8 @@ pub struct BottomUpCheckpointManager<T> {
     metadata: CheckpointConfig,
     parent_handler: T,
     child_handler: T,
+    /// The number of blocks away from the chain head that is considered final
+    finalization_blocks: ChainEpoch,
```
adlrocha, line 32: As described above, I don't think we need this parameter (at least for Tendermint). We can leave it here, but let's make the default 0

ipc/provider/src/checkpoint.rs
```diff
@@ -151,21 +169,41 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
     async fn submit_next_epoch(&self, submitter: &Address) -> Result<()> {
         let next_submission_height = self.next_submission_height().await?;
```
adlrocha, line 170: Just a sanity-check: remember that relayers are allowed to submit a checkpoint for height `h` even if it has been submitted already as long as no submission have been made yet for `h+1`

ipc/cli/src/commands/checkpoint/relayer.rs
cryptoAtwill: Currently it is default to 0 if user does not set it.

ipc/provider/src/checkpoint.rs
cryptoAtwill: yep, there is a submit last method

PR #361, opened 2023-11-01
IPC-347: add pre-fund command and genesis balances in GenesisInfo call

adlrocha:
`ipc-cli` implementation for https://github.com/consensus-shipyard/fendermint/issues/347

Depends on: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/267

This PR:
* Adds a flag to include genesis balance for a validator when it joins. 
* Adds a `pre-fund` command for users to be able to add genesis balance in a child subnet before it is bootsrapped.
* It includes the genesis balance information in the `GenesisInfo` call made through the provider so Fendermint can use it to generate its genesis with this initial balance information.
* Adds a `pre-release` command to recover funds from the initial balance of a subnet before it bootstraps.

### DIFF
--- a/ipc/cli/src/commands/subnet/join.rs
+++ b/ipc/cli/src/commands/subnet/join.rs
@@ -179,13 +179,13 @@ impl CommandLineHandler for PreFundSubnet {
 #[derive(Debug, Args)]
 #[command(
     name = "pre-fund",
-    about = "Pre fund with some funds in genesis in a child-subnet"
+    about = "Add some funds in genesis to an address in a child-subnet"
 )]
 pub struct PreFundSubnetArgs {
     #[arg(long, short, help = "The address funded in the subnet")]
     pub from: Option<String>,
     #[arg(long, short, help = "The subnet to add balance to")]
     pub subnet: String,
-    #[arg(help = "Optionally add an initial balance to the validator in genesis in the subnet")]
+    #[arg(help = "Add an initial balance for the address in genesis in the subnet")]
     pub initial_balance: f64,
 }
--- a/ipc/cli/src/commands/subnet/join.rs
+++ b/ipc/cli/src/commands/subnet/join.rs
@@ -32,7 +32,7 @@ impl CommandLineHandler for JoinSubnet {
         if let Some(initial_balance) = arguments.initial_balance {
             println!("pre-funding address with {initial_balance}");
             provider
-                .prefund_subnet(subnet.clone(), from, f64_to_token_amount(initial_balance)?)
+                .pre_fund(subnet.clone(), from, f64_to_token_amount(initial_balance)?)
                 .await?;
         }
         let epoch = provider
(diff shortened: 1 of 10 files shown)


aakoshh:
I'll have a closer look, but I'm slightly surprised there's a new `pre-fund` command instead of just using `fund`, so I thought I'd ask: is there any way for removing a balance from a subnet that isn't getting enough collateral to be activated? For example validators can call `leave` and get their tokens back immediately if the subnet isn't bootstrapped. I realise `release` is supposed to be called on the subnet chain, so I'm not sure how something like this fits in.

ipc/cli/src/commands/subnet/join.rs
```diff
@@ -37,7 +43,7 @@ impl CommandLineHandler for JoinSubnet {
                 public_key,
             )
             .await?;
-        log::info!("joined at epoch: {epoch}");
+        println!("joined at epoch: {epoch}");
```
aakoshh, line 46: Why stop using `log`?

ipc/cli/src/commands/subnet/join.rs
```diff
@@ -136,3 +147,45 @@ pub struct UnstakeSubnetArgs {
...
+#[command(
+    name = "pre-fund",
+    about = "Pre fund with some funds in genesis in a child-subnet"
+)]
+pub struct PreFundSubnetArgs {
+    #[arg(long, short, help = "The address funded in the subnet")]
+    pub from: Option<String>,
+    #[arg(long, short, help = "The subnet to add balance to")]
+    pub subnet: String,
+    #[arg(help = "Optionally add an initial balance to the validator in genesis in the subnet")]
```
aakoshh, line 189: It's not optional here. Can only validators use this?

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -312,6 +312,27 @@ impl SubnetManager for EthSubnetManager {
         block_number_from_receipt(receipt)
     }
 
+    async fn pre_fund(&self, subnet: SubnetID, from: Address, balance: TokenAmount) -> Result<()> {
+        let balance = balance
+            .atto()
+            .to_u128()
+            .ok_or_else(|| anyhow!("invalid min validator stake"))?;
```
aakoshh, line 319:
```suggestion
            .ok_or_else(|| anyhow!("invalid fund amount"))?;
```

ipc/cli/src/commands/subnet/join.rs
```diff
@@ -136,3 +147,45 @@ pub struct UnstakeSubnetArgs {
...
+            None => None,
+        };
+        provider
+            .prefund_subnet(
+                subnet.clone(),
+                from,
+                f64_to_token_amount(arguments.initial_balance)?,
+            )
+            .await?;
+        println!("address pre-funded successfully");
```
aakoshh, line 173:
```suggestion
        log::info!("address pre-funded successfully");
```

aakoshh (commented):
Looks good, my only question is what I asked above: how do you get your balance back if things don't go according to plan?

ipc/cli/src/commands/subnet/join.rs
adlrocha: Mikhail suggested that command output should ascribe to the Unix UX recommendations. That means that command outputs shouldn't return logs but raw prints so they can be piped and processed easily.

ipc/cli/src/commands/subnet/join.rs
adlrocha: Same as above, we are moving from logs in command returns.

adlrocha:
> Looks good, my only question is what I asked above: how do you get your balance back if things don't go according to plan?

I initially thought this wouldn't be an issue but now that you mention it, it definitely is a big one. I will merge this one and add a `pre-release` method in the contract and its corresponding command here to allow any user to get their funds back before the subnet bootstraps in case they need to.

ipc/cli/src/commands/subnet/join.rs
adlrocha: Typo due to copy-past. Fixed! Thanks!

ipc/provider/src/manager/evm/manager.rs
adlrocha: Fixed. Copy-pasta again. Thanks!

cryptoAtwill (approved):

ipc/cli/src/commands/subnet/join.rs
aakoshh: I don't think this is a real return value, not something you would want to pipe into another program. The only return value from this should be a 0 exit code. There is no variable data in it.

PR #379, opened 2023-11-06
IPC-373: Fix balances returned with bogus rpc

### LINKED ISSUE #373: wallet balances returns even if bogus RPC
### Issue type

Bug

Yes

### Version

dev

### Custom code

Yes

_No response_

### Describe the issue

`./bin/ipc-cli wallet balances` returns no error even if using a bad (non-existent) RPC in the config; it just returns 0 balances for every wallet. The expected behaviour would be to return an error.

### Repro steps

Run `./bin/ipc-cli wallet balances --subnet XXX --wallet-type evm` using a well-formed but bogus RPC endpoint in the ipc-cli config.

### Relevant log output

```shell
# Correct RPC
ubuntu@ip-172-26-12-88:~/ipc$ ./bin/ipc-cli wallet balances --subnet /r314159/t410fqypdmnt3tyfczkis6iobfgqq72yk4web4ytehdi --wallet-type evm
0x310c8812089ee8efce8b06302ecf58a5c0fe549b - Balance: 1.0
0xc3964d4b507cb3dc4f24a5f0962fd7339ea568a7 - Balance: 0.0
0x93ef94dc6bda3283263366e569b0a0ffda1677b7 - Balance: 1.0
0x5126aa09cde4d5d85fc080cc3c069a8098eb2e10 - Balance: 1.0
default-key - Balance: 0

# Bogus RPC
ubuntu@ip-172-26-12-88:~/ipc$ ./bin/ipc-cli wallet balances --subnet /r314159/t410fqypdmnt3tyfczkis6iobfgqq72yk4web4ytehdi --wallet-type evm
default-key - Balance: 0
0xc3964d4b507cb3dc4f24a5f0962fd7339ea568a7 - Balance: 0
0x5126aa09cde4d5d85fc080cc3c069a8098eb2e10 - Balance: 0
0x310c8812089ee8efce8b06302ecf58a5c0fe549b - Balance: 0
0x93ef94dc6bda3283263366e569b0a0ffda1677b7 - Balance: 0
```

adlrocha:
Fixes #373

### DIFF
--- a/ipc/cli/src/commands/wallet/balances.rs
+++ b/ipc/cli/src/commands/wallet/balances.rs
@@ -30,7 +30,6 @@ impl CommandLineHandler for WalletBalances {
             WalletType::Evm => {
                 let wallet = provider.evm_wallet()?;
                 let addresses = wallet.read().unwrap().list()?;
-                let mut no_balance = addresses.clone();
                 let r = addresses
                     .iter()
                     .map(|addr| {
@@ -50,15 +49,18 @@ impl CommandLineHandler for WalletBalances {
 
                 let v: Vec<anyhow::Result<(TokenAmount, &EthKeyAddress)>> = join_all(r).await;
 
-                for r in v.into_iter().filter_map(|r| r.ok()) {
-                    let (balance, addr) = r;
-                    if addr.to_string() != "default-key" {
-                        println!("{} - Balance: {}", addr.to_string(), balance);
-                        no_balance.retain(|a| a != addr);
+                for r in v.into_iter() {
+                    match r {
+                        Ok(i) => {
+                            let (balance, addr) = i;
+                            if addr.to_string() != "default-key" {
+                                println!("{} - Balance: {}", addr.to_string(), balance);
+                            }
+                        }
+                        Err(e) => {
+                            return Err(e);
+                        }
                     }
-                }
-                for addr in no_balance {
-                    println!("{} - Balance: 0", addr.to_string());
                 }
             }
             WalletType::Fvm => {
--- a/ipc/cli/src/commands/wallet/balances.rs
+++ b/ipc/cli/src/commands/wallet/balances.rs
@@ -26,6 +26,8 @@ impl CommandLineHandler for WalletBalances {
 
         let wallet_type = WalletType::from_str(&arguments.wallet_type)?;
         let subnet = SubnetID::from_str(&arguments.subnet)?;
+        let mut errors = Vec::new();
+
         match wallet_type {
             WalletType::Evm => {
                 let wallet = provider.evm_wallet()?;
@@ -58,9 +60,18 @@ impl CommandLineHandler for WalletBalances {
                             }
                         }
                         Err(e) => {
-                            return Err(e);
+                            errors.push(e);
                         }
                     }
+                }
+
+                if !errors.is_empty() {
+                    let error = errors
+                        .into_iter()
+                        .fold(anyhow::anyhow!("Error fetching balances"), |acc, err| {
+                            acc.context(err)
+                        });
+                    return Err(error);
                 }
             }
             WalletType::Fvm => {


ipc/cli/src/commands/wallet/balances.rs
```diff
@@ -50,16 +49,19 @@ impl CommandLineHandler for WalletBalances {
...
-                    let (balance, addr) = r;
-                    if addr.to_string() != "default-key" {
-                        println!("{} - Balance: {}", addr.to_string(), balance);
-                        no_balance.retain(|a| a != addr);
+                for r in v.into_iter() {
+                    match r {
+                        Ok(i) => {
+                            let (balance, addr) = i;
+                            if addr.to_string() != "default-key" {
+                                println!("{} - Balance: {}", addr.to_string(), balance);
```
aakoshh, line 57: I know we talked about this, but `jsonline` would be a good output format IMO, as it would make it easy to do something with the result using `jq`. A human readable output format is a great option, but personally I think it's liberating to print out structured data an defer the responsibility of formatting to the consumer.

ipc/cli/src/commands/wallet/balances.rs
```diff
@@ -50,16 +49,19 @@ impl CommandLineHandler for WalletBalances {
...
-                for r in v.into_iter().filter_map(|r| r.ok()) {
-                    let (balance, addr) = r;
-                    if addr.to_string() != "default-key" {
-                        println!("{} - Balance: {}", addr.to_string(), balance);
-                        no_balance.retain(|a| a != addr);
+                for r in v.into_iter() {
+                    match r {
+                        Ok(i) => {
+                            let (balance, addr) = i;
+                            if addr.to_string() != "default-key" {
```
aakoshh, line 56: What is this `"default-key"`?

ipc/cli/src/commands/wallet/balances.rs
```diff
@@ -50,16 +49,19 @@ impl CommandLineHandler for WalletBalances {
...
+                                println!("{} - Balance: {}", addr.to_string(), balance);
+                            }
+                        }
+                        Err(e) => {
+                            return Err(e);
```
aakoshh, line 61: Okay so we went from ignoring errors with `.filter_map(|r| r.ok())` and printing 0 balance for them to printing the first few successful balances until we encounter an error, where we stop going down the list and return an overall failure. 

That seems a bit inconsistent, that depending on which of the N requests failed I see more or less results, compared to for example:
* Print all results if all of them succeeded or none if any of them failed
* Print all results as JSON regardless of error or failure, but if it failed then print eg.. `{addr: "...", error: "oh noes!"}`
* Print all the successful results first and then return an overall error if any failed

The all-or-nothing strategy can be achieved for example like so:

```rust
let v: Vec<(TokenAmount, &EthKeyAddress)> = join_all(r).await.collect::<anyhow::Result<Vec<_>>>()?;
```

ipc/cli/src/commands/wallet/balances.rs
adlrocha: I kind of agree, but let's tackle this as a joint effort after LabWeek and let's track the final approach we want to follow here: https://github.com/consensus-shipyard/ipc/issues/365. So far depending on who is reviewing the PR they suggest one approach or the other :see_no_evil:

For this one specifically, this approach seems more readable for humans than returning a JSON.

ipc/cli/src/commands/wallet/balances.rs
adlrocha: This is the internal key used by the wallet store to tag the address being used as the default one for the wallet.

ipc/cli/src/commands/wallet/balances.rs
adlrocha: This one is a bit tricky, because there are errors for which all balances will fail with the same error (like RPC unavailable), and others that may be address-specific (like address not initialized on-chain). I agree that my approach may have not been the best because it was non-deterministic, so I decided to go with your suggestion of printing all of the balances that succeed, and then returning the errors (if any) at the end. That way if it this is a balance-wide error (like RPC) we return a single error, and if it is address-specific we print the balances and then the errors. Let me know what you think.

ipc/cli/src/commands/wallet/balances.rs
aakoshh: It's strange to see that returned in the API :thinking: 
And why do we skip printing the balance? Aren't you interested in seeing the balance of your default (maybe only) address? In JSON this could be `{addr: "f410...", balance: 1.23, default: true}`

ipc/cli/src/commands/wallet/balances.rs
aakoshh: I'm sure it is more readable, and I agree you can postpone it. I guess like `ls -lh` you could have a "human readable hint" but it could also end up being an overkill.

ipc/cli/src/commands/wallet/balances.rs
```diff
@@ -50,15 +51,27 @@ impl CommandLineHandler for WalletBalances {
...
                 }
-                for addr in no_balance {
-                    println!("{} - Balance: 0", addr.to_string());
+
+                if !errors.is_empty() {
+                    let error = errors
+                        .into_iter()
+                        .fold(anyhow::anyhow!("Error fetching balances"), |acc, err| {
+                            acc.context(err)
+                        });
```
aakoshh, line 73: Nice

aakoshh (approved):

ipc/cli/src/commands/wallet/balances.rs
adlrocha: ^ I like this idea of having a flag for human-readable v.s. computer-readable. Will track it in the issue :pray:

ipc/cli/src/commands/wallet/balances.rs
jsoares: I believe the default still stands as its own entry, no? At least I still see my default address printed out.

jsoares (approved):
It seems to do the job, i.e. now it errors out in case of bad RPC and prints fine in case of good RPC.

PR #383, opened 2023-11-06
FIX: remove docker-build related code from install_infra

adlrocha:
Remove `docker-build` related reference from the code. `install_infra` should only pull `fendermint` to get the latest `cargo make` recipes for the infrastructure.


jsoares (approved):
Tested, works fine.

PR #397, opened 2023-11-15
Update quickstart-calibration.md

hammertoe:
Fix node id command name


jsoares (approved):
It's unclear to me that we're still maintaining the github quickstart, but change looks good.

PR #402, opened 2023-11-21
Topdown query by events

cryptoAtwill:
Query topdown messages through event. This needs to be merged after: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/285

### DIFF
--- a/ipc/provider/src/manager/evm/manager.rs
+++ b/ipc/provider/src/manager/evm/manager.rs
@@ -6,11 +6,11 @@ use std::collections::{BTreeMap, HashMap};
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
 
-use ethers::types::H256;
 use ethers_contract::{ContractError, EthLogDecode, LogMeta};
 use ipc_actors_abis::{
     gateway_getter_facet, gateway_manager_facet, gateway_messenger_facet, gateway_router_facet,
-    lib_staking_change_log, subnet_actor_getter_facet, subnet_actor_manager_facet, subnet_registry,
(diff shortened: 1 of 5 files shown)


ipc/provider/src/manager/evm/manager.rs
```diff
@@ -116,31 +116,42 @@ impl TopDownFinalityQuery for EthSubnetManager {
...
+            .from_block(epoch as u64)
+            .to_block(epoch as u64)
+            .topic1(topic1);
+
+        let mut messages = vec![];
+        for (event, meta) in query_with_meta(ev, gateway_contract.client()).await? {
+            let msg = CrossMsg::try_from(event.message)?;
+            log::debug!("received message: {:?} and meta: {:?}", msg, meta);
+
+            if meta.block_hash.0 != block_hash {
```
adlrocha, line 143: It shouldn't be the case that a block hash is different for one of the events of the block, right? So if the block has a different hash it should error in the first iteration. 

Also, we should de-duplicate events here in case there are reorgs. Remember the issue we found when there was a reorg and we were fetching duplicate events with the `removed` flag.  I'll try to find the PR where Akosh addresses this.

cryptoAtwill: I think it's checking the first iteration as well? Probably because I did the parsing first (so that I can log whatever before the error), that gave the wrong impression?
Akosh did it in the `query_with_meta` function, so it should be save here.

adlrocha: right, you are using `query_with_meta` not `ev.query_with_meta`.

adlrocha (approved):
LGTM. Feel free to merge.

PR #404, opened 2023-11-27
align topdown and validator changes query pattern

cryptoAtwill:
Align the query pattern of validator change set and top down messages, so that their querying pattern is consistent, easier to maintain. 

Instead of passing in a block hash to check for top down messages query (we had to do this because previously we were using getters to do the querying), we are querying top down messages with block hash returned from events and let caller do the check (we can do this now because we are using events). This is aligned with validator change set query pattern and it's more flexible.

### DIFF
--- a/ipc/provider/src/manager/evm/manager.rs
+++ b/ipc/provider/src/manager/evm/manager.rs
@@ -110,12 +110,7 @@ impl TopDownFinalityQuery for EthSubnetManager {
         &self,
         subnet_id: &SubnetID,
         epoch: ChainEpoch,
-        block_hash: &[u8],
-    ) -> Result<Vec<CrossMsg>> {
-        if block_hash.len() != 32 {
-            return Err(anyhow!("invalid block hash len"));
-        }
-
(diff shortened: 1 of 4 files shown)


ipc/provider/src/manager/evm/manager.rs
```diff
@@ -136,22 +130,28 @@ impl TopDownFinalityQuery for EthSubnetManager {
...
-            let msg = CrossMsg::try_from(event.message)?;
-            log::debug!("received message: {:?} and meta: {:?}", msg, meta);
-
-            if meta.block_hash.0 != block_hash {
-                return Err(anyhow!(
-                    "block hash not equal, input: {}, received: {}",
-                    hex::encode(block_hash),
-                    hex::encode(meta.block_hash.0)
-                ));
+            if let Some(h) = hash {
```
adlrocha, line 135: Is this check still needed?

adlrocha (approved):

ipc/provider/src/manager/evm/manager.rs
cryptoAtwill: yeah, we have to ensure the events have the same block hashes.

PR #405, opened 2023-11-29
Fix prefund

cryptoAtwill:
Fix prefund by calling `fund` if `prefund` fails.

This should fix https://github.com/consensus-shipyard/ipc-monorepo/issues/353 and https://github.com/consensus-shipyard/ipc-monorepo/issues/357


adlrocha (changes requested):
I don't think the current implementation fully solve the issues linked, would you mind having a look at my comments? Thanks!

PR #412, opened 2023-12-05
IPC-304: add permissioned flag for new contract version

adlrocha:
Depends on: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/305
Upgrades contract version and introduces a `permissioned` flag to address https://github.com/consensus-shipyard/ipc-solidity-actors/issue/304

### DIFF
--- a/ipc/cli/src/commands/subnet/create.rs
+++ b/ipc/cli/src/commands/subnet/create.rs
@@ -41,6 +41,7 @@ impl CreateSubnet {
                     .active_validators_limit
                     .unwrap_or(DEFAULT_ACTIVE_VALIDATORS),
                 f64_to_token_amount(arguments.min_cross_msg_fee)?,
+                arguments.permissioned,
             )
             .await?;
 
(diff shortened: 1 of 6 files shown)


ipc/cli/src/commands/subnet/create.rs
```diff
@@ -96,4 +97,9 @@ pub struct CreateSubnetArgs {
         help = "Minimum fee for cross-net messages in subnet (in whole FIL)"
     )]
     pub min_cross_msg_fee: f64,
+    #[arg(
+        long,
```
maciejwitowski, line 101: the conversion to `long` will be done automatically by the CLI library we use, right?

maciejwitowski (approved):
Approved, with comments.

ipc/cli/src/commands/subnet/create.rs
adlrocha: This `long` what specifies is that the cli will only expect the long form of the flag name i.e. `--permissioned` not the short version which would be `-p`.

ipc/cli/src/commands/subnet/create.rs
```diff
@@ -96,4 +97,9 @@ pub struct CreateSubnetArgs {
...
+        help = "Deploy static network where validators can't join in a permissionless manner"
+    )]
+    pub permissioned: bool,
```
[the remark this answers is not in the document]
adlrocha: Good point, we should probably make it optional, being `false` the default value.

PR #414, opened 2023-12-07
Update mycelium address

jsoares:
Updates Mycelium address following restart on 06/12.


jsoares:
This needs to be propagated to `main` so that it shows for visitors.

maciejwitowski (approved):

PR #415, opened 2023-12-20
Sync with latest solidity

cryptoAtwill:
Since the `solidity` contracts are updated, a lot, there will be a series of PRs to sync with latest solidity contracts:

- [x] Subnet lifecycle: 
  - changes the `create` subnet and other methods to align with the solidity contract. 
  - Added a lot of data types as well as removing some.
  - **Commented off** bottom up related code, especially the relayer related temporarily so that the code compile while spending the time to understand the new bottom up checkpoint submission flow.
- [ ] Bottom up checkpoint related: TODO
- [ ] Relayer: TODO


aakoshh (approved):
I don't know the details either, but in the interest of moving this along, :+1: from me.

